PR #8, opened 2016-10-20
JENKINS-39145 - adding ability to pass actions list beside the parameters list

yurigo79:
resolving https://issues.jenkins-ci.org/browse/JENKINS-39145


jglick:
Do not plan to do this, and we did, the implementation would likely be different.

PR #12, opened 2017-03-06
[JENKINS-38887] Allow triggering ComputedFolders and other Queue.Task implementations

stephenc:
- Cannot wait for non-Job tasks to complete though

See [JENKINS-38887](https://issues.jenkins-ci.org/browse/JENKINS-38887)

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -39,6 +39,7 @@
     </pluginRepositories>
     <properties>
         <jenkins.version>1.642.3</jenkins.version>
+        <scm-api.version>2.0.7</scm-api.version>
     </properties>
     <dependencies>
         <dependency>
@@ -103,13 +104,13 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>scm-api</artifactId>
-            <version>2.0.7</version>
+            <version>${scm-api.version}</version>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>scm-api</artifactId>
-            <version>2.0.7</version>
+            <version>${scm-api.version}</version>
             <classifier>tests</classifier>
             <scope>test</scope>
         </dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -94,5 +94,24 @@
             <version>2.1</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>branch-api</artifactId>
+            <version>2.0.7</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>scm-api</artifactId>
+            <version>2.0.7</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>scm-api</artifactId>
+            <version>2.0.7</version>
+            <classifier>tests</classifier>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 </project>
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStep.java
@@ -133,7 +133,7 @@ public class BuildTriggerStep extends AbstractStepImpl {
         }
 
         public AutoCompletionCandidates doAutoCompleteJob(@AncestorInPath ItemGroup<?> container, @QueryParameter final String value) {
-            // TODO remove code copy&pasted from AutoCompletionCandidates.ofJobNames when it supports union of classes
+            // TODO remove code copy&pasted from AutoCompletionCandidates.ofJobNames when it supports testing outside Item bound
             final AutoCompletionCandidates candidates = new AutoCompletionCandidates();
             class Visitor extends ItemVisitor {
                 String prefix;
@@ -154,7 +154,7 @@ public class BuildTriggerStep extends AbstractStepImpl {
                             && i.hasPermission(Item.READ)
                         // and read permission required
                             ) {
-                        if ((i instanceof ParameterizedJobMixIn.ParameterizedJob || i instanceof Queue.Task) && n.startsWith(value))
+                        if (i instanceof Queue.Task && n.startsWith(value))
                             candidates.add(n);
 
                         // recurse
@@ -218,7 +218,7 @@ public class BuildTriggerStep extends AbstractStepImpl {
 
         public FormValidation doCheckJob(@AncestorInPath ItemGroup<?> context, @QueryParameter String value) {
             if (StringUtils.isBlank(value)) {
-                return FormValidation.ok();
+                return FormValidation.warning(Messages.BuildTriggerStep_no_job_configured());
             }
             Item item = Jenkins.getActiveInstance().getItem(value, context, Item.class);
             if (item == null) {
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStep.java
@@ -2,11 +2,16 @@ package org.jenkinsci.plugins.workflow.support.steps.build;
 
 import hudson.Extension;
 import hudson.model.AutoCompletionCandidates;
+import hudson.model.Describable;
+import hudson.model.Item;
 import hudson.model.ItemGroup;
+import hudson.model.ItemVisitor;
 import hudson.model.Job;
 import hudson.model.ParameterDefinition;
 import hudson.model.ParameterValue;
 import hudson.model.ParametersDefinitionProperty;
+import hudson.model.Queue;
+import hudson.model.TopLevelItem;
 import hudson.util.FormValidation;
 import java.util.ArrayList;
 import java.util.List;
@@ -14,6 +19,7 @@ import jenkins.model.Jenkins;
 import jenkins.model.ParameterizedJobMixIn;
 import net.sf.json.JSONArray;
 import net.sf.json.JSONObject;
+import org.apache.commons.lang.StringUtils;
 import org.jenkinsci.plugins.workflow.steps.AbstractStepDescriptorImpl;
 import org.jenkinsci.plugins.workflow.steps.AbstractStepImpl;
 import org.jenkinsci.plugins.workflow.steps.Step;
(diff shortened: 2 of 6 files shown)


pom.xml
```diff
@@ -94,5 +94,24 @@
...
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>branch-api</artifactId>
+            <version>2.0.7</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>scm-api</artifactId>
+            <version>2.0.7</version>
```
jglick: introduce a POM property

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepExecution.java
```diff
@@ -53,39 +58,83 @@
...
-            LOGGER.log(Level.FINER, "scheduling a build of {0} from {1}", new Object[] {project, context});
-        }
-        actions.add(new CauseAction(new Cause.UpstreamCause(invokingRun)));
-        List<ParameterValue> parameters = step.getParameters();
-        if (parameters != null) {
-            parameters = completeDefaultParameters(parameters, (Job) project);
-            actions.add(new ParametersAction(parameters));
+        if (step.getWait() && !(item instanceof Job)) {
+            // TODO find some way of allowing ComputedFolders to hook into the listener code
+            throw new AbortException("Waiting for non-job items is not supported");
```
jglick: Note that this kind of condition could also be checked in `doCheck*` methods for the benefit of **Pipeline Syntax**.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepExecution.java
```diff
@@ -53,39 +58,83 @@
...
-        // TODO use new convenience method in 1.621
-        if (quietPeriod == null) {
-            quietPeriod = project.getQuietPeriod();
-        }
-        QueueTaskFuture<?> f = new ParameterizedJobMixIn() {
-            @Override protected Job asJob() {
-                return (Job) project;
+        if (item instanceof ParameterizedJobMixIn.ParameterizedJob) {
+            final ParameterizedJobMixIn.ParameterizedJob project = (ParameterizedJobMixIn.ParameterizedJob) item;
+            listener.getLogger().println("Scheduling project: " + ModelHyperlinkNote.encodeTo(project));
```
jglick: [Reduced diff](https://github.com/jenkinsci/pipeline-build-step-plugin/pull/12/files?w=1#diff-a8a21e28a85973ee663505521435630cR69)

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepExecution.java
```diff
@@ -53,39 +58,83 @@
...
+            }
+            actions.add(new CauseAction(new Cause.UpstreamCause(invokingRun)));
+            Integer quietPeriod = step.getQuietPeriod();
+            if (quietPeriod == null) {
+                try {
+                    Method getQuietPeriod = task.getClass().getMethod("getQuietPeriod");
+                    if (getQuietPeriod.getReturnType().equals(int.class)) {
+                        quietPeriod = (Integer) getQuietPeriod.invoke(task);
+                    }
+                } catch (NoSuchMethodException | IllegalAccessError | IllegalArgumentException | InvocationTargetException e) {
```
jglick: Only `NoSuchMethodException` should be ignored; other exceptions should be logged.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepExecution.java
```diff
@@ -53,39 +58,83 @@
...
+                    // ignore, best effort only
+                }
+            }
+            if (quietPeriod == null) {
+                quietPeriod = Jenkins.getActiveInstance().getQuietPeriod();
+            }
+            ScheduleResult scheduleResult = Jenkins.getActiveInstance().getQueue().schedule2(task, quietPeriod, actions);
+            if (scheduleResult.isRefused() || scheduleResult.getItem() == null) {
```
jglick: Checking `getItem` is redundant according to `isRefused` Javadoc.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepTest.java
```diff
@@ -359,4 +375,42 @@ public void cancelBuildQueue() throws Exception {
...
+        @Override
+        public boolean recognizes(@NonNull ItemGroup<?> parent, @NonNull String name,
+                                  @NonNull List<? extends SCMSource> scmSources,
+                                  @NonNull Map<String, Object> attributes,
+                                  @CheckForNull SCMHeadEvent<?> event, @NonNull TaskListener listener)
+                throws IOException, InterruptedException {
+            return false;
+        }
+
+        @TestExtension
```
jglick:
```java
("triggerOrgFolder")
```

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepExecution.java
```diff
@@ -53,39 +58,83 @@
...
+            }
+            Queue.Task task = (Queue.Task) item;
+            node.addAction(new LabelAction(Messages.BuildTriggerStepExecution_building_(task.getFullDisplayName())));
+            List<Action> actions = new ArrayList<>();
+            if (step.getWait()) {
+                StepContext context = getContext();
+                actions.add(new BuildTriggerAction(context, step.isPropagate()));
+                LOGGER.log(Level.FINER, "scheduling a build of {0} from {1}", new Object[]{task, context});
+            }
+            actions.add(new CauseAction(new Cause.UpstreamCause(invokingRun)));
```
jglick: Actually could use `singletonList`.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepExecution.java
```diff
@@ -53,39 +58,83 @@
...
             }
-        }.scheduleBuild2(quietPeriod, actions.toArray(new Action[actions.size()]));
-        if (f == null) {
-            throw new AbortException("Failed to trigger build of " + project.getFullName());
+        } else if (item instanceof Queue.Task){
+            if (step.getParameters() != null && !step.getParameters().isEmpty()) {
+                throw new AbortException("Item type does not support parameters");
+            }
+            Queue.Task task = (Queue.Task) item;
+            node.addAction(new LabelAction(Messages.BuildTriggerStepExecution_building_(task.getFullDisplayName())));
```
jglick: Any reason you are not printing a `ModelHyperlinkNote` in this mode?

pom.xml
stephenc: Coincidentally the same version

stephenc: Oh sorry you mean for scm-api

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepExecution.java
stephenc: Then I need to check or assert that `getItem()!=null` or face the wrath of findbugs (kann's lesser known brother)

pom.xml
jglick: Yes, sorry, would have been clearer to comment on the second usage.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepExecution.java
jglick: Sorry, why? You are not using the result of `getItem()` that I can see.

> Review is a dish best served cold.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepExecution.java
stephenc: done

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStep.java
```diff
@@ -143,5 +202,39 @@ public FormValidation doCheckPropagate(@QueryParameter boolean value, @QueryPara
...
+            }
+            if (item instanceof Job) {
+                return FormValidation.ok();
+            }
+            return FormValidation.error(Messages.BuildTriggerStep_no_wait_for_non_jobs());
+        }
+
+        public FormValidation doCheckJob(@AncestorInPath ItemGroup<?> context, @QueryParameter String value) {
+            if (StringUtils.isBlank(value)) {
+                return FormValidation.ok();
```
jglick: Not sure what the standard UI is for initially blank but mandatory fields.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepExecution.java
```diff
@@ -102,29 +103,26 @@ protected Job asJob() {
                 throw new AbortException("Item type does not support parameters");
             }
             Queue.Task task = (Queue.Task) item;
+            listener.getLogger().println("Scheduling item: " + ModelHyperlinkNote.encodeTo(item));
             node.addAction(new LabelAction(Messages.BuildTriggerStepExecution_building_(task.getFullDisplayName())));
-            List<Action> actions = new ArrayList<>();
-            if (step.getWait()) {
-                StepContext context = getContext();
-                actions.add(new BuildTriggerAction(context, step.isPropagate()));
```
jglick: Well. There could be a `Job & Queue.Task` which is not a `ParameterizedJobMixIn.ParameterizedJob`. I do not know of any examples (`ExternalJob` is not schedulable) so this is hypothetical, but we could wind up in a situation where we accept `wait` and such a job, and so the step starts and goes to sleep but then will never wake up on its own.

(Yes I know I said you could use `singletonList`; I forgot about the `BuildTriggerAction`.)

stephenc:
@jglick 
0fc20831a2773edf17b80ffb74ff09bc

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStep.java
```diff
@@ -180,9 +180,9 @@ private String contextualNameOf(Item i) {
                 if (value.startsWith("/"))
                     new Visitor("/").onItemGroup(Jenkins.getInstance());
 
-                for (String p = "../"; value.startsWith(p); p += "../") {
+                for (StringBuilder p = new StringBuilder("../"); value.startsWith(p.toString()); p .append("../")) {
```
jglick: probably should be patched in core too

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStep.java
```diff
@@ -143,5 +202,39 @@ public FormValidation doCheckPropagate(@QueryParameter boolean value, @QueryPara
...
+        public FormValidation doCheckJob(@AncestorInPath ItemGroup<?> context, @QueryParameter String value) {
+            if (StringUtils.isBlank(value)) {
+                return FormValidation.warning(Messages.BuildTriggerStep_no_job_configured());
+            }
+            Item item = Jenkins.getActiveInstance().getItem(value, context, Item.class);
+            if (item == null) {
+                return FormValidation.error(Messages.BuildTriggerStep_cannot_find(value));
+            }
+            if (item instanceof Job) {
+                return FormValidation.ok();
```
jglick: This is not OK if it does not implement `Queue.Task`, right?

PR #13, opened 2017-06-27
Save information about child jobs in parent job

Jimilian:
In some cases it would  be nice to know what exactly was triggered by pipeline job. I.e. this information can be used by Build Failure Analyzer Plugin to show failures in child jobs on the top trigger job.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildInfoAction.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildInfoAction.java
@@ -6,6 +6,7 @@ import jenkins.model.Jenkins;
 import hudson.model.InvisibleAction;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Queue;
 import java.util.concurrent.ConcurrentLinkedQueue;
@@ -22,7 +23,11 @@ public class BuildInfoAction extends InvisibleAction {
     }
 
     public List<Run<?, ?>> getChildBuilds() {
-        List<Run<?, ?>> builds = new ArrayList<>();
+        if (buildInfos.isEmpty()) {
+            return Collections.emptyList();
+        }
+
+        List<Run<?, ?>> builds = new ArrayList<>(buildInfos.size());
 
         for (BuildInfo buildInfo : buildInfos) {
             if (buildInfo != null) {
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildInfoAction.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildInfoAction.java
@@ -1,40 +1,56 @@
 package org.jenkinsci.plugins.workflow.support.steps.build;
 
+import hudson.model.Job;
+import hudson.model.Run;
+import jenkins.model.Jenkins;
 import hudson.model.InvisibleAction;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Queue;
+import java.util.concurrent.ConcurrentLinkedQueue;
 
 
 public class BuildInfoAction extends InvisibleAction {
-    private final List<BuildInfo> buildInfoList = new ArrayList<>();
+    private final Queue<BuildInfo> buildInfos = new ConcurrentLinkedQueue<>();
 
-    public BuildInfoAction(String projectName, int buildNumber) {
-        buildInfoList.add(new BuildInfo(projectName, buildNumber));
+    BuildInfoAction(String projectName, int buildNumber) {
+        buildInfos.add(new BuildInfo(projectName, buildNumber));
     }
 
-    public List<BuildInfo> getBuildInfoList() {
-        return buildInfoList;
+    public List<Run> getChildBuilds() {
+        List<Run> builds = new ArrayList<>();
+
+        for (BuildInfo buildInfo : buildInfos) {
+            if (buildInfo != null) {
+                Job job = Jenkins.getActiveInstance().getItemByFullName(buildInfo.getProjectName(), Job.class);
+                if (buildInfo.getBuildNumber() != 0) {
+                    builds.add((job != null) ? job.getBuildByNumber(buildInfo.getBuildNumber()) : null);
+                }
+            }
+        }
+
+        return builds;
     }
 
     public void addBuildInfo(String projectName, int buildNumber) {
-        buildInfoList.add(new BuildInfo(projectName, buildNumber));
+        buildInfos.add(new BuildInfo(projectName, buildNumber));
     }
 
-    static class BuildInfo {
+    private static class BuildInfo {
         private final String projectName;
         private final int buildNumber;
 
-        public BuildInfo(String projectName, int buildNumber) {
+        BuildInfo(String projectName, int buildNumber) {
             this.projectName = projectName;
             this.buildNumber = buildNumber;
         }
 
-        public String getProjectName() {
+        String getProjectName() {
             return projectName;
         }
 
-        public int getBuildNumber() {
+        int getBuildNumber() {
             return buildNumber;
         }
     }
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildInfoAction.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildInfoAction.java
@@ -14,12 +14,15 @@ import java.util.concurrent.ConcurrentLinkedQueue;
 public class BuildInfoAction extends InvisibleAction {
     private final Queue<BuildInfo> buildInfos = new ConcurrentLinkedQueue<>();
 
+    BuildInfoAction() {
+    }
+
     BuildInfoAction(String projectName, int buildNumber) {
         buildInfos.add(new BuildInfo(projectName, buildNumber));
     }
 
-    public List<Run> getChildBuilds() {
-        List<Run> builds = new ArrayList<>();
+    public List<Run<?, ?>> getChildBuilds() {
+        List<Run<?, ?>> builds = new ArrayList<>();
 
         for (BuildInfo buildInfo : buildInfos) {
             if (buildInfo != null) {
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildInfoAction.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildInfoAction.java
@@ -1,39 +1,30 @@
 package org.jenkinsci.plugins.workflow.support.steps.build;
 
+import hudson.model.InvisibleAction;
 import hudson.model.Job;
 import hudson.model.Run;
 import jenkins.model.Jenkins;
-import hudson.model.InvisibleAction;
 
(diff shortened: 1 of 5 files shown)


src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildInfoAction.java
```diff
@@ -0,0 +1,41 @@
+package org.jenkinsci.plugins.workflow.support.steps.build;
+
+import hudson.model.InvisibleAction;
+
+import java.util.ArrayList;
+import java.util.List;
+
+
+public class BuildInfoAction extends InvisibleAction {
```
oleg-nenashev: I see no problem with making it visible, but maybe it's outside the scope

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildInfoAction.java
```diff
@@ -0,0 +1,41 @@
...
+    private final List<BuildInfo> buildInfoList = new ArrayList<>();
+
+    public BuildInfoAction(String projectName, int buildNumber) {
+        buildInfoList.add(new BuildInfo(projectName, buildNumber));
+    }
+
+    public List<BuildInfo> getBuildInfoList() {
+        return buildInfoList;
```
oleg-nenashev: Wrap with unmodifyableList()

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildInfoAction.java
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    public void addBuildInfo(String projectName, int buildNumber) {
+        buildInfoList.add(new BuildInfo(projectName, buildNumber));
```
oleg-nenashev: It seems there may be concurrency issues with this field. I would recommend CopyOnWrite list or `synchronize`

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildInfoAction.java
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    static class BuildInfo {
+        private final String projectName;
+        private final int buildNumber;
+
+        public BuildInfo(String projectName, int buildNumber) {
```
oleg-nenashev: This approach may start leaking job existence fact if the API user has no `Item.Discover` permission. Not sure if it's acceptable

Jimilian: How can I prevent this?
Btw, is it possible that user has access to list of actions of some build, but doesn't have access to the log of this build? (right now plugin publishes names/build numbers of triggered jobs to the log).

For now I made this class private and only list of runs can be retrieved from action. I hope that it would be safer.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildInfoAction.java
Jimilian: CopyOnWrite looks like overkill, `synchronize` needs some manual work, so, I picked up `ConcurrentLinkedQueue` (current implementation of this action doesn't require random access).

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildInfoAction.java
Jimilian: Gone.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildInfoAction.java
Jimilian: I can put this comment to the code ;)

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerListener.java
```diff
@@ -29,6 +30,15 @@ public void onStarted(Run<?, ?> run, TaskListener listener) {
                     TaskListener taskListener = stepContext.get(TaskListener.class);
                     // encodeTo(Run) calls getDisplayName, which does not include the project name.
                     taskListener.getLogger().println("Starting building: " + ModelHyperlinkNote.encodeTo("/" + run.getUrl(), run.getFullDisplayName()));
+                    FlowNode parentNode = stepContext.get(FlowNode.class);
+                    if (parentNode != null) {
```
Jimilian: @oleg-nenashev, I suppose that this code is not thread-safe as well. Is it correct?
I think to put `synchronized(this)` for `addAction/getAction`

Jimilian:
@oleg-nenashev, @jglick, can you take a look, please?

Jimilian:
@jglick, will you have time to review this change? Should I rebase it first?

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildInfoAction.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    BuildInfoAction(String projectName, int buildNumber) {
+        buildInfos.add(new BuildInfo(projectName, buildNumber));
+    }
+
+    public List<Run<?, ?>> getChildBuilds() {
+        List<Run<?, ?>> builds = new ArrayList<>();
+
+        for (BuildInfo buildInfo : buildInfos) {
```
oleg-nenashev: Minor optimization: Return `Collections.emptyList()` if the size is null before creating a new ArrayList. Maybe also makes sense to pre-initialize `builds` size to buildInfos.size

oleg-nenashev: Not sure if `ConcurrentLinkedQueue.size()` is fast enough though. Maybe it will hammer the performance

Jimilian: `ConcurrentLinkedQueue.isEmpty()` comes for free, but in current implementation it's not possible to have empty collection of `buildInfos` longer than several ms (+ StopTheWorld pause :) ). Anyway, I added this check anyway - who knows how this `Action` will be used in future.
Pre-initialize - good idea, can decrease memory footprint a little bit.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerListener.java
```diff
@@ -29,6 +30,11 @@ public void onStarted(Run<?, ?> run, TaskListener listener) {
...
```
jglick: No need to check for null if defined via `@StepContextParameter` (which is actually deprecated but fixing that requires a larger change).

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerListener.java
```diff
@@ -38,6 +44,20 @@ public void onStarted(Run<?, ?> run, TaskListener listener) {
         }
     }
 
+    private BuildInfoAction ensureBuildInfoAction(final FlowNode parentNode) {
+        BuildInfoAction buildInfoAction = parentNode.getAction(BuildInfoAction.class);
+        if (buildInfoAction == null) {
+            synchronized (this) {
+                buildInfoAction = parentNode.getAction(BuildInfoAction.class);
+                if (buildInfoAction == null) {
```
jglick: Double-checked locking, unsafe.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildInfoAction.java
```diff
@@ -0,0 +1,65 @@
...
+        return builds;
+    }
+
+    public void addBuildInfo(String projectName, int buildNumber) {
+        buildInfos.add(new BuildInfo(projectName, buildNumber));
+    }
+
+    private static class BuildInfo {
+        private final String projectName;
+        private final int buildNumber;
```
jglick: Easier just to use `String` via `Run.externalizableId`.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepTest.java
```diff
@@ -430,6 +450,23 @@ public MultiBranchProjectFactory newInstance() {
         }
     }
 
+    private void assertBuildInfoAction(WorkflowRun workflowRun, String projectName, int buildNumber) {
+        FlowGraphWalker walker = new FlowGraphWalker(workflowRun.getExecution());
```
svanoort: 🐜  Deprecated, use DepthFirstScanner instead

svanoort:
I'll more or less echo Jesse's comments here -- would need to see an example of how this API is used in practice.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepTest.java
Jimilian: It needs `workflow-api:2.17`, so, I updated it in `pom.xml` as well, but now it fails in verification: https://ci.jenkins.io/blue/organizations/jenkins/Plugins%2Fpipeline-build-step-plugin/detail/PR-13/9/tests/
Also, I think @jglick would not be happy to introduce dependency for higher version (especially in case of test code needs it -> not production).

So, what should I do?

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildInfoAction.java
Jimilian: It does not look easier for me, in this case I would need one String concatenation + `String.split` for every call for nothing. I now that it's tiny stuff, but even from code perspective it doesn't simplify much.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildInfoActionTest.java
```diff
@@ -0,0 +1,81 @@
...
+
+        while (true) {
+            if (future1.isDone() && future2.isDone() && future3.isDone() && future4.isDone() && future5.isDone() ) {
+                break;
+            }
+
+            Thread.sleep(100);
+        }
+
+        assertEquals(501, buildInfoAction.getChildBuilds().size());
```
Jimilian: Right now I used `PowerMock` only for Jenkins singleton, for the rest of stuff I used Mockito. What do you prefer to use for mocking Jenkins.getInstance method?

Jimilian:
@svanoort, @jglick I wanted to use it Build Failure Analyzer Plugin (@rsandell). I created proof of concept some time ago and it works perfect, but to create good PR I need to submit this change first. So, I wanted to use API presented in this PR in this way: https://github.com/Jimilian/build-failure-analyzer-plugin/blob/pipeline_jobs_poc/src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/dbf/BuildFlowDBF.java

PR #14, opened 2017-08-24
JENKINS-46311 Show las 1000 lines of logs from Downstream jobs.

roidelapluie:
I am looking for help & guidance for this pull request.


roidelapluie:
Hi @jglick yes that is what I want. I just have no idea how I can do it yet.

roidelapluie:
@jglick Hi...

Improved version...

- still no tests
- race condition (last second of the logs is not printed)

roidelapluie:
but there is streaming now

Dmitry1987:
Hi guys, was this abandoned?  Is there a way to pull logs from downstream jobs?

jglick:
Abandoned. No opt-in/out, no truncation, no support for controller restarts, etc. At this point I doubt we would even want something like this; better to enhance pipeline visualizations to better navigate to downstream builds.

PR #18, opened 2017-12-12
Mark downstream runs with pointer to upstream FlowNode

sophistifunk:
Adds a `BuildUpstreamNodeAction` to downstream jobs referring to the triggering node, so that Blue Ocean can annotate the correct pipeline step when viewing results.

See: https://issues.jenkins-ci.org/browse/JENKINS-38339


sophistifunk:
I'll create a PR for the Blue Ocean side of this tomorrow, and link to it here also.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepExecution.java
```diff
@@ -6,33 +6,9 @@
...
-import hudson.model.Job;
-import hudson.model.ParameterDefinition;
-import hudson.model.ParameterValue;
-import hudson.model.ParametersAction;
-import hudson.model.ParametersDefinitionProperty;
-import hudson.model.Queue;
-import hudson.model.Result;
-import hudson.model.Run;
-import hudson.model.TaskListener;
+import hudson.model.*;
```
jglick: Avoid wildcard imports please.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildUpstreamNodeAction.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.Objects;
+
+/**
+ * Attached to newly-created builds in order to point back to the triggering FlowNode.
+ *
+ * We annotate the downstream build instead of upstream in order to support no-wait builds,
+ * because the downstream run id is not available when they're still queued.
+ *
+ * Needed for Blue Ocean to annotate the correct step.
+ * @see: https://issues.jenkins-ci.org/browse/JENKINS-38339
```
jglick: invalid tag syntax

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepExecution.java
```diff
@@ -65,18 +51,23 @@ public boolean start() throws Exception {
             // TODO find some way of allowing ComputedFolders to hook into the listener code
             throw new AbortException("Waiting for non-job items is not supported");
         }
+
+        List<Action> actions = new ArrayList<>();
+        actions.add(new CauseAction(new Cause.UpstreamCause(invokingRun)));
+        actions.add(new BuildUpstreamNodeAction(node, invokingRun));
```
jglick: Or I suppose you could simply subclass `UpstreamCause` to add the `FlowNode.id` information.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildUpstreamNodeAction.java
```diff
@@ -0,0 +1,53 @@
...
+import hudson.model.Run;
+import org.jenkinsci.plugins.workflow.graph.FlowNode;
+
+import java.util.Objects;
+
+/**
+ * Attached to newly-created builds in order to point back to the triggering FlowNode.
+ *
+ * We annotate the downstream build instead of upstream in order to support no-wait builds,
+ * because the downstream run id is not available when they're still queued.
```
jglick: The `Queue.Item.id` _is_ available FWIW, so you _could_ place the marker on the upstream build if that were the more convenient direction to retrieve later. (You cannot jump directly from a `Queue.Item.id` to a `Run.number`, if the queue item was in fact successfully scheduled and started building; you can only check the queue ID of a given build.)

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildUpstreamNodeAction.java
```diff
@@ -0,0 +1,53 @@
...
+    public String getUpstreamNodeId() {
+        return upstreamNodeId;
+    }
+
+    public String getUpstreamRunId() {
+        return upstreamRunId;
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
jglick: Probably unnecessary to override `equals` & `hashCode`.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildUpstreamNodeAction.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    public String getUpstreamNodeId() {
+        return upstreamNodeId;
+    }
+
+    public String getUpstreamRunId() {
+        return upstreamRunId;
+    }
+
```
jglick: Might want to add a convenience method to return the actual upstream `FlowNode`, if the upstream build still exists etc. Depends on what your downstream PR looks like.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepExecution.java
sophistifunk: Ack. Will change this.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepExecution.java
sophistifunk: It seems like this would make things more complicated elsewhere, without much benefit besides saving a few hundred bytes in cold storage.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildUpstreamNodeAction.java
sophistifunk: This would be more code and much more work for the server so I think we should leave it as-is, but thanks for pointing it out, I didn't know the queue had its own id space for jobs.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildUpstreamNodeAction.java
sophistifunk: Wouldn't really help much in this case, as BO doesn't actually return all FlowNodes back via REST, and we have to walk the graph a little to annotate the node that makes it out to the UI. And that code may change once I've talked to Vivek about it, because I have a suspicion it (the BO code) is not the best way to do what we need.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildUpstreamNodeAction.java
sophistifunk: I'll change this also

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildUpstreamNodeAction.java
sophistifunk: I like to be on the safe side when I know something is going in and out of collections, and there are places in the API docs that specifically mention making sure you have `equals` and `hashCode` on `Action`s

sophistifunk:
Related BO pull request here: https://github.com/jenkinsci/blueocean-plugin/pull/1587

PR #19, opened 2017-12-20
[FIXED JENKINS-48632] Ensure descriptions are included on downstream parameters

abayer:
[JENKINS-48632](https://issues.jenkins-ci.org/browse/JENKINS-48632)

cc @reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -38,10 +38,10 @@
         </pluginRepository>
     </pluginRepositories>
     <properties>
-        <jenkins.version>1.642.3</jenkins.version>
+        <jenkins.version>2.7.3</jenkins.version>
         <java.level>7</java.level>
-        <scm-api.version>2.0.7</scm-api.version>
-        <workflow-step-api-plugin.version>2.13</workflow-step-api-plugin.version>
+        <scm-api.version>2.0.8</scm-api.version>
+        <workflow-step-api-plugin.version>2.14</workflow-step-api-plugin.version>
     </properties>
     <dependencies>
         <dependency>
@@ -52,12 +52,17 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-api</artifactId>
-            <version>2.3</version>
+            <version>2.22</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-support</artifactId>
-            <version>2.6</version>
+            <version>2.14</version>
+        </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>script-security</artifactId>
+            <version>1.39</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -38,10 +38,10 @@
         </pluginRepository>
     </pluginRepositories>
     <properties>
-        <jenkins.version>1.642.3</jenkins.version>
+        <jenkins.version>2.7.3</jenkins.version>
```
abayer: Various plugin/core bumps are so we can do `properties(...)` and `parameters: [booleanParam(...)]` and have fairly current (for core 2.7.3, at least) dependencies for tests. I basically lifted all the dependency versions straight from Declarative, which is also on 2.7.3.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepExecution.java
```diff
@@ -151,26 +152,27 @@ protected Job asJob() {
...
-                    if (!names.contains(pDef.getName())) {
+                    if (!allParameters.containsKey(pDef.getName())) {
                         ParameterValue defaultP = pDef.getDefaultParameterValue();
                         if (defaultP != null) {
-                            completeListOfParameters.add(defaultP);
+                            allParameters.put(defaultP.getName(), defaultP);
                         }
+                    } else {
+                        // Get the description of specified parameters here. UI submission of parameters uses formatted description.
+                        allParameters.get(pDef.getName()).setDescription(pDef.getFormattedDescription());
```
jglick: That does not accord with what I see [here](https://github.com/jenkinsci/jenkins/blob/90f778d2819cfd223f8815a3d84ce3e904d91fb0/core/src/main/java/hudson/model/StringParameterDefinition.java#L129) (web submission) or [here](https://github.com/jenkinsci/jenkins/blob/90f778d2819cfd223f8815a3d84ce3e904d91fb0/core/src/main/java/hudson/model/StringParameterDefinition.java#L134) (`build` CLI command submission). And I see no test coverage of `description`s that would be transformed by a `MarkupFormatter`.

As of SECURITY-353. [`ParameterValue.description` is not assumed to have been formatted](https://github.com/jenkinsci/jenkins/blob/90f778d2819cfd223f8815a3d84ce3e904d91fb0/core/src/main/resources/hudson/model/StringParameterValue/value.jelly#L30).

src/test/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepTest.java
```diff
@@ -477,4 +479,37 @@ public MultiBranchProjectFactory newInstance() {
...
+    @Test
+    public void parameterDescriptions() throws Exception {
+        WorkflowJob ds = j.jenkins.createProject(WorkflowJob.class, "ds");
+        ds.setDefinition(new CpsFlowDefinition("properties([\n" +
+                "  parameters([\n" +
+                "    booleanParam(defaultValue: true, description: 'flag description', name: 'flag'),\n" +
+                "    string(defaultValue: 'default string', description: 'strParam description', name: 'strParam')\n" +
+                "  ])\n" +
+                "])\n", true));
+        // Define the parameters
```
jglick: easier to add the `ParametersDefinitionProperty` programmatically

src/test/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepTest.java
```diff
@@ -477,4 +479,37 @@ public MultiBranchProjectFactory newInstance() {
...
+
+        WorkflowRun r = ds.getBuildByNumber(2);
+        assertNotNull(r);
+
+        ParametersAction action = r.getAction(ParametersAction.class);
+        assertNotNull(action);
+
+        ParameterValue flagValue = action.getParameter("flag");
+        assertNotNull(flagValue);
+        assertEquals("flag description", flagValue.getDescription());
```
jglick: IIUC this was the one assertion that failed otherwise?

jglick:
> As of SECURITY-353

Perhaps you should bump your baseline to [2.32.2](https://jenkins.io/security/advisory/2017-02-01/) or newer, so that you are testing against what users should actually be running, rather than something which became officially obsolete well over a year ago.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepExecution.java
abayer: Works for me.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepTest.java
abayer: Correct.

PR #23, opened 2019-04-12
[FIXED JENKINS-52038] Fail for invalid choice parameter values

abayer:
[JENKINS-52038](https://issues.jenkins-ci.org/browse/JENKINS-52038)


jglick:
> [the build step] accepts parameters that are not defined

As designed. [Discussion](https://jenkins.io/security/advisory/2016-05-11/#arbitrary-build-parameters-are-passed-to-build-scripts-as-environment-variables)

jglick:
> until that API trickles down into a sufficiently old/usable LTS

So where is that work being done? If it is not done now, it is not going to happen at all. I think the core API should be done first, the use of it in this plugin prepared using a JEP-305 version, and only _then_ this released as a temporary workaround.

PR #25, opened 2019-06-13
Document default behaviour for propagate field in build step

ovidiub13:


src/main/resources/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStep/help-propagate.html
```diff
@@ -2,4 +2,5 @@
     If set, then if the downstream build is anything but successful (blue ball), this step fails.
```
dwnusbaum: nit: I'd tweak this line instead and delete the new line you added.

```suggestion
     If enabled (it is enabled by default), then if the downstream build is anything but successful (blue ball), this step fails.
```

PR #26, opened 2019-06-13
Add information about the returned build object

ovidiub13:


src/main/resources/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStep/help.html
```diff
@@ -1,3 +1,4 @@
 <div>
     Triggers a new build for a given job.
+    Returns a <code>build</code> object that can be used to retrieve information about the build. Object is documented in your Jenkins instance, under <code>/pipeline-syntax/globals#currentBuild</code>.
```
dwnusbaum: If possible, I think it would be ideal to link directly to the page with an `<a>` tag rather than just showing what page has the necessary information.

ovidiub13: How can I link to the users Jenkins instance?

dwnusbaum: I think that if you use a relative URL everything should just work, right? Maybe not in the case of reverse proxies and similar?

ovidiub13: I can't use a relative URL since the user might read theis from the web documentation, like I did, not from the built in docs.

dwnusbaum: Ah I wasn't thinking about the web documentation, good point. Maybe link to the Javadoc for [RunWrapper](https://javadoc.jenkins.io/plugin/workflow-support/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.html), and reword to something like this:
```suggestion
    Returns a <code>RunWrapper</code> instance for the triggered build. Documentation for RunWrapper is available <a href="https://javadoc.jenkins.io/plugin/workflow-support/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.html">here</a>, or in your Jenkins instance under <code>/pipeline-syntax/globals#currentBuild</code>.
```

papajulio (approved):

xgin:
I was looking for exactly this information! It would be great to have it in the documentation.

petamas:
I'd also add it to the documentation that the return value is `null` if you set the `wait` parameter to `false`. (If I understand the code correctly.)

madpipeline:
@dwnusbaum Can we merge this?

PR #27, opened 2019-07-12
Migrate snippet generator tests from workflow-cps

dwnusbaum:
It looks like tests pass and the snippet generator works correctly even without this override, so I'm not sure what it's for. If it turns out it is necessary, I'd like to add a test and/comment explaining what it is here for.

EDIT: In the end, just migrating some tests from workflow-cps, see https://github.com/jenkinsci/workflow-cps-plugin/pull/304.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStep.java
@@ -86,39 +86,6 @@ public class BuildTriggerStep extends AbstractStepImpl {
             super(BuildTriggerStepExecution.class);
         }
 
-        @Override public Step newInstance(StaplerRequest req, JSONObject formData) throws FormException {
-            BuildTriggerStep step = (BuildTriggerStep) super.newInstance(req, formData);
-            // Cf. ParametersDefinitionProperty._doBuild:
-            Object parameter = formData.get("parameter");
-            JSONArray params = parameter != null ? JSONArray.fromObject(parameter) : null;
-            if (params != null) {
-                Job<?,?> context = StaplerReferer.findItemFromRequest(Job.class);
-                Job<?,?> job = Jenkins.getActiveInstance().getItem(step.getJob(), context, Job.class);
-                if (job != null) {
-                    ParametersDefinitionProperty pdp = job.getProperty(ParametersDefinitionProperty.class);
-                    if (pdp != null) {
-                        List<ParameterValue> values = new ArrayList<ParameterValue>();
-                        for (Object o : params) {
-                            JSONObject jo = (JSONObject) o;
-                            String name = jo.getString("name");
-                            ParameterDefinition d = pdp.getParameterDefinition(name);
-                            if (d == null) {
-                                throw new IllegalArgumentException("No such parameter definition: " + name);
-                            }
-                            ParameterValue parameterValue = d.createValue(req, jo);
-                            if (parameterValue != null) {
-                                values.add(parameterValue);
-                            } else {
-                                throw new IllegalArgumentException("Cannot retrieve the parameter value: " + name);
-                            }
-                        }
-                        step.setParameters(values);
-                    }
-                }
-            }
-            return step;
-        }
-
         @Override
         public String getFunctionName() {
             return "build";
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStep.java
@@ -86,6 +86,43 @@ public class BuildTriggerStep extends AbstractStepImpl {
             super(BuildTriggerStepExecution.class);
         }
 
+        // Note: This is necessary because the JSON format of the parameters produced by config.jelly when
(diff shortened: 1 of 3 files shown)


jglick:
I think https://github.com/jenkinsci/workflow-cps-plugin/blob/609bc51d4cca65e8c851cdbc05a3f0efd8d60638/src/test/java/org/jenkinsci/plugins/workflow/cps/SnippetizerTest.java#L203-L238 are the relevant tests. Possibly I forgot to move them from https://github.com/jenkinsci/pipeline-plugin/pull/69 + https://github.com/jenkinsci/pipeline-plugin/pull/190 during the 2.x split. `workflow-cps` generates a test JAR so `SnippetizerTester` can be used from anywhere.

dwnusbaum:
Now I'm extra confused, because [`SnippitizerTest.buildTriggerTest`](https://github.com/jenkinsci/workflow-cps-plugin/blob/6ac97e69c03896463016dee58b8deb6b4a56d420/src/test/java/org/jenkinsci/plugins/workflow/cps/SnippetizerTest.java#L171) seems to pass with these changes. Maybe I my SNAPSHOT versions are messed up locally.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepTest.java
```diff
@@ -549,4 +553,16 @@ public void invalidChoiceParameterValue() throws Exception {
...
                 j.assertBuildStatus(Result.FAILURE, us.scheduleBuild2(0)));
     }
+
+    @Test public void buildTriggerStep() throws Exception {
+        SnippetizerTester st = new SnippetizerTester(j);
+        BuildTriggerStep step = new BuildTriggerStep("downstream");
+        st.assertRoundTrip(step, "build 'downstream'");
+        step.setParameters(Arrays.asList(new StringParameterValue("branch", "default"), new BooleanParameterValue("correct", true)));
+        if (StringParameterDefinition.DescriptorImpl.class.isAnnotationPresent(Symbol.class)) {
+            st.assertRoundTrip(step, "build job: 'downstream', parameters: [string(name: 'branch', value: 'default'), booleanParam(name: 'correct', value: true)]");
```
dwnusbaum: Ok, the reason this passes but the actual UI doesn't work is that `Snippetizer.assertRoundTrip` doesn't test that snippet generation roundtrips correctly, only that an already-configured Step roundtrips with the given text when converted to Groovy (i.e. DSL.invokeStep). You could delete config.jelly entirely, so that the snippet generator UI was totally broken for the step, and `SnippetizerTester.assertRoundTrip` wouldn't care. In our case, the problem is in configuring the step from the UI, but that is not tested.

dwnusbaum:
> I think https://github.com/jenkinsci/workflow-cps-plugin/blob/609bc51d4cca65e8c851cdbc05a3f0efd8d60638/src/test/java/org/jenkinsci/plugins/workflow/cps/SnippetizerTest.java#L203-L238 are the relevant tests

Ah, whoops, I missed those, was only looking at `SnippetizerTest.buildTriggerStep`, and they definitely catch the issue. I'll move them all here.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepTest.java
jglick: I think this is resolved by https://github.com/jenkinsci/pipeline-build-step-plugin/pull/27#issuecomment-511021550 right?

dwnusbaum: Yes, closing.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStep.java
```diff
@@ -86,6 +86,9 @@ public DescriptorImpl() {
             super(BuildTriggerStepExecution.class);
         }
 
+        // Note: This is necessary because the JSON format of the parameters produced by config.jelly when
+        // using the snippet generator does not match what would be neccessary for databinding to work automatically.
+        // For non-snippet generator use, this is unnecessary.
         @Override public Step newInstance(StaplerRequest req, JSONObject formData) throws FormException {
```
jglick: BTW you could update the `structs` dep and use `CustomDescribableModel`.

dwnusbaum: Is there something that ties `Descriptor.newInstance(StaplerRequest, JSONObject)` to `DescribableModel`, or do you mean implement `CustomDescribableModel` and replace the body here with something like the following?

```
   return DescribableModel.of(BuildTriggerStep.class).instantiate(convertJsonToMap(formData))
```

Also, unless it is safe to assume that data binding for `ParameterDefinition`s via `DescribableModel` will operate identically to `ParameterDefinition.createValue(StaplerRequest, JSONObject)` we'd still have to use that method in `customInstantiate`, which would be a bit awkward.

Seems easiest just to leave it for now.

jglick: Sorry, forgot this was the overload with `JSONObject`. Ignore.

PR #33, opened 2019-09-17
Test propagation of additional downstream result types

basil:
#24 changed the behavior of this plugin to propagate the result of downstream builds. However, only propagation of unstable builds was covered in unit tests, and the documentation was not updated. This PR adds test coverage for the other types of downstream statuses and updates the documentation.

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStep/help-propagate.html
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStep/help-propagate.html
@@ -1,5 +1,6 @@
 <p>
-    If enabled (default state), then if the downstream build is anything but successful (blue ball), this step fails.
+    If enabled (default state), then the result of this step is that of the downstream build (e.g.,
+    success, unstable, failure, not built, or aborted).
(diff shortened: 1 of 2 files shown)


basil:
FYI, I think #24 is going in the right direction, but it also makes me a bit nervous, because I'm not sure all of the `parallel` steps in my pipelines are going to correctly deal with the downgrade of build results as a result of this change in behavior. I may find some regressions in my jobs and have to introduce `propagate: false` and propagate the results from `parallel` steps manually instead. I expressed my use case in more detail in [JENKINS-49073 (comment)](https://issues.jenkins-ci.org/browse/JENKINS-49073?focusedCommentId=375702&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-375702). As I described in that comment, I think allowing the `parallel` step to propagate the "worst" result from a set of parallel branches would be very helpful.

basil:
@dwnusbaum Gentle ping. No rush on this, but just wanted to make sure the first notification didn't slip by.

dwnusbaum:
@basil Sorry, it didn't slip by, I've just been swamped lately with other work. Not sure when I'll be able to take a look through open PRs.

PR #34, opened 2019-10-04
Add auto-conversion of parameters

Wadeck:
Currently if you do not configure correctly your "parent" pipeline, you could by mistake trigger a child job while enforcing an expected PasswordParameterValue to become a StringParameterValue.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepExecution.java
@@ -24,6 +24,8 @@ import hudson.model.PasswordParameterValue;
 import hudson.model.Queue;
 import hudson.model.Result;
 import hudson.model.Run;
+import hudson.model.SimpleParameterDefinition;
+import hudson.model.StringParameterDefinition;
 import hudson.model.StringParameterValue;
 import hudson.model.TaskListener;
 import hudson.model.queue.QueueTaskFuture;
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepExecution.java
```diff
@@ -167,15 +171,31 @@ protected Job asJob() {
...
                         }
                     } else {
+                        String description = Util.fixNull(pDef.getDescription());
                         if (pDef instanceof ChoiceParameterDefinition) {
                             ParameterValue pv = allParameters.get(pDef.getName());
                             if (!((ChoiceParameterDefinition)pDef).getChoices().contains(pv.getValue())) {
                                 throw new AbortException("Value for choice parameter '" + pDef.getName() + "' is '" + pv.getValue() + "', "
                                         + "but valid choices are " + ((ChoiceParameterDefinition)pDef).getChoices());
                             }
+                        } else if (pDef instanceof PasswordParameterDefinition) {
```
dwnusbaum: Why not just convert every `StringParameterValue` where the `ParameterDefinition` is a subtype of `SimpleParameterDefinition` but not `StringParameterDefinition` using `SimpleParameterDefinition.createValue` as in https://github.com/jenkinsci/parameterized-trigger-plugin/pull/62?

Wadeck: I tried your proposal, please look at the last commit. From my PoV it adds complexity due to the number of corner cases.

Feel free to revert the last commit, change it yourself, or merge it like this :)

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepExecution.java
```diff
@@ -167,15 +173,37 @@ protected Job asJob() {
...
                         }
                     } else {
+                        String description = Util.fixNull(pDef.getDescription());
                         if (pDef instanceof ChoiceParameterDefinition) {
                             ParameterValue pv = allParameters.get(pDef.getName());
                             if (!((ChoiceParameterDefinition)pDef).getChoices().contains(pv.getValue())) {
                                 throw new AbortException("Value for choice parameter '" + pDef.getName() + "' is '" + pv.getValue() + "', "
                                         + "but valid choices are " + ((ChoiceParameterDefinition)pDef).getChoices());
                             }
+                        } else if (pDef instanceof StringParameterDefinition) {
```
dwnusbaum: FWIW I meant to not even special-case passwords at all, just handle all `SimpleParameterDefinition`s the same way, something like this:

```
if (pDef instanceof SimpleParameterDefinition && !pDef instanceof StringParameterDefinition) {
    ParameterValue pv = allParameters.get(pDef.getName());
    if (pv instanceof StringParameterValue) {
        String desiredValue = (String) pv.getValue();
        listener.getLogger().println(String.format("The types of the specified parameters did not match the expected types defined on the target job. Converting '%s' to satisfy '%s'.", pv.getName(), pDef.getClass().getSimpleName()));
        ParameterValue convertedValue = ((SimpleParameterDefinition) pDef).createValue(desiredValue);
        allParameters.put(pDef.getName(), convertedValue);
    }
}
```
To me that seems ideal because it means that pipeline-build-step and parameterized-trigger have the exact same behavior, which seems least likely to cause confusion.

I like the idea of throwing `AbortException` if the parameter types do not match and cannot be converted, but parameterized-trigger does not seem to do that, so I am a little hesitant to be more strict here, and I have no idea of whether that case would occur in practice and might break things, or if it should never happen at all.

What do you think? Not trying to force one way or the other, just trying to understand the problem space better.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepExecution.java
dwnusbaum: Replied down in https://github.com/jenkinsci/pipeline-build-step-plugin/pull/34#discussion_r333023820, mostly I am just trying to understand the problem better, not saying "do this" or "do that". Thanks for submitting a PR for this!

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepExecution.java
dwnusbaum: Also, let me know if you want me to take this over, not sure how much time you have to spend on it.

Wadeck: It was just to speed-up the reaction by providing the initial investigation, unit tests and a potentially working solution. So if you can take over the task, it will be with pleasure :) 

Concerning the special cases, it's because the parameterized-trigger and this plugin does not share the same situation IIUC. Within a pipeline, you cannot pass any arbitrary parameter, like `choiceParam` are passed as `string`. Perhaps it's just due to my minimum knowledge of the context.

The only case that *must* be covered is when you pass a string and the job expected a password, all the other behaviors are up to you ;)

madchap:
@dwnusbaum do you need help from me to the get the necessary time? 
cc @vivek this is rotting.

dwnusbaum:
Closing in favor of #35.

PR #35, opened 2019-11-18
[JENKINS-60216] Make the build step automatically convert some kinds of parameter mismatches

dwnusbaum:
See [JENKINS-60216](https://issues.jenkins-ci.org/browse/JENKINS-60216). Subsumes #34. 

Tries to keep the behavior as similar to https://github.com/jenkinsci/parameterized-trigger-plugin/pull/62 as possible, since that behavior has been around for a long time so it seems unlikely to be problematic (there was a [followup commit](https://github.com/jenkinsci/parameterized-trigger-plugin/commit/8f410be6a8a203e92c4120a12750514878585869) for some issues, but from reading the PR and associated tickets, it seems that case only applied to Choice Parameters, which are already special-cased here and will not be converted by this PR).

I'll create a JIRA ticket for this tomorrow for tracking.

@Wadeck: I added you as a co-author to the commit because it is based on a squashed version of your PR, but let me know if you want me to remove you as a co-author and I will do that.

Open questions:

* [ ] Should we add a system property to enable/disable this behavior?

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepTest.java
@@ -19,8 +19,10 @@ import hudson.model.Label;
 import hudson.model.ParameterValue;
 import hudson.model.ParametersAction;
 import hudson.model.ParametersDefinitionProperty;
+import hudson.model.PasswordParameterDefinition;
 import hudson.model.Queue;
 import hudson.model.Result;
+import hudson.model.Run;
 import hudson.model.StringParameterDefinition;
 import hudson.model.StringParameterValue;
 import hudson.model.TaskListener;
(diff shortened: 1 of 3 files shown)


src/test/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepTest.java
```diff
@@ -595,4 +598,49 @@ public void invalidChoiceParameterValue() throws Exception {
...
+        // A password parameter passed as a text parameter is converted.
+        us.setDefinition(new CpsFlowDefinition(String.format(def, "text(name: 'password', value: 'secret')"), true));
+        WorkflowRun b2 = j.buildAndAssertSuccess(us);
+        j.assertLogContains("The parameter 'password' did not have the type expected", b2);
+        assertThat(getParameter(ds.getBuildByNumber(2), "password").getDescription(), equalTo(Messages.BuildTriggerStepExecution_convertedParameterDescription("description", "Password Parameter", us.getName() + " #2")));
+
+        // A password parameter passed as a password parameter is not converted.
+        us.setDefinition(new CpsFlowDefinition(String.format(def, "password(name: 'password', value: 'secret')"), true));
+        WorkflowRun b3 = j.buildAndAssertSuccess(us);
+        j.assertLogNotContains("The parameter 'password' did not have the type expected", b3);
```
dwnusbaum: TODO: I want to add a test where a password parameter is passed as a boolean parameter to show that the boolean parameter is not converted.

dwnusbaum: Done in a049a76.

PR #38, opened 2019-12-02
JENKINS-60345: Add trigger build link

soenkekueper:
Added an new pipeline step that allows to add an customizable link on the build's page to start an downstream link after user interaction.


jglick:
Interesting but out of scope for this plugin. See discussion in https://issues.jenkins.io/browse/JENKINS-60345

PR #42, opened 2020-01-15
[JENKINS-60779] Do not log conversion warnings for choice-like parameter definitions

dwnusbaum:
See [JENKINS-60779](https://issues.jenkins-ci.org/browse/JENKINS-60779).

Not sure about adding tests for this, I'd really prefer not to have a dependency on either of the relevant plugins, even in test scope.


dwnusbaum:
The Windows agent is having issues, rebuilding:

> win2019-741a30 was marked offline: Connection was broken: java.io.EOFException

pault1337:
It would make sense to add the following classes "org.biouno.unochoice.CascadeChoiceParameter" and "org.biouno.unochoice.ChoiceParameter" in the condition. If you're using the Active Choice Plugin, you'll have the warning otherwise though it should be the same exclusion as Extensible Choice Parameter.

PR #43, opened 2020-03-17
Add changelog entry for 2.12

dwnusbaum:
I am planning on releasing this plugin today.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 ## Changelog
 
+### 2.12
+
+Release date: 2020-03-17
+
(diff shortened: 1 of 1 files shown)


dwnusbaum:
> This looks fine, but I'm wondering if we can switch to use GitHub releases instead of changelog file?

We could. Personally, I don't like the format of the releases page for my use cases, so I haven't really been motivated to start using release drafter. 🤷‍♂

dwnusbaum:
CI failure is because of an infra issue:

> EC2 (aws) - Ubuntu 18.04 LTS (i-0d7eecb72d3a4806a) was marked offline: Connection was broken: java.io.EOFException

Given that this PR only modifies the changelog, and the last build of master is green, I am going to go ahead and merge this.

PR #45, opened 2020-07-06
Use new isValid() API for ParameterDefinition.

jeffret-b:
This demonstrates how to use the new `ParameterDefinition.isValid` method instead of implementing the logic locally. This is related to the new API in core from jenkinsci/jenkins#4826. See [JENKINS-62889](https://issues.jenkins-ci.org/browse/JENKINS-62305). This enhancement is expected to first appear in Jenkins 2.245 and this PR cannot be merged until that change is incorporated into all supported versions.

The test failure is related to [JENKINS-62889](https://issues.jenkins-ci.org/browse/JENKINS-62305).


jglick:
Somehow relates to #46 but I do not follow the context. @dwnusbaum do we still want this PR? Do we need some other sort of cleanup?

jglick:
@dwnusbaum do you recall what this was about, and whether it is still useful?

dwnusbaum:
I do not remember it. It seems to be cleanup to use a new API from https://github.com/jenkinsci/jenkins/pull/4826 instead of duplicating the check logic.

jglick (approved):
:shrug: the `AbortException.message` is not as nice. OTOH this could potentially handle other `ParameterDefinition.isValid` overrides. Seems there is still some tech debt related to #42, I guess because the API design for parameters has some fundamental flaws in terms of introspection.

PR #46, opened 2020-07-24
[JENKINS-62305] Make password parameters work with the build step in Jenkins 2.236+

dwnusbaum:
See [JENKINS-62305](https://issues.jenkins-ci.org/browse/JENKINS-62305).

https://github.com/jenkinsci/jenkins/pull/4630 changed the `@DataBoundConstructor` for `PasswordParameterValue` so that one of its parameters is a `Secret`, but `Secret` is not supported by data binding via `DescribableModel` in `structs`. 

This PR uses `CustomDescribableModel` to maintain compatibility with existing uses of the `build` step with password parameters, and adds documentation explaining that credentials should be preferred over password parameters going forward.

TODO: Do we want to add a runtime warning (e.g. message in the build log) any time a password parameter is used with the `build` step? My understanding is that while password parameters are not ideal, they are not completely broken, so I am a little hesitant to add a warning if the issues are relatively minor.

Other things that are broken and that this PR does not attempt to fix:
* Snippet generation for password parameters. I am not sure exactly what is wrong, but the new UI control for passwords from https://github.com/jenkinsci/jenkins/pull/3991 seems to be non-interactive in the snippet generator. Additionally, the generated Groovy snippet shows `<object of type hudson.util.Secret>`, so we'd need to implement `customUninstantiate` to fix that.
* Doc generation for the `build` step (https://www.jenkins.io/doc/pipeline/steps/pipeline-build-step/). Once the job that creates those docs is updated to run against Jenkins 2.236 or higher, it will fail. A special case will need to be added to https://github.com/jenkins-infra/pipeline-steps-doc-generator to handle `PasswordParameterValue`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,24 @@
 
 Adds the Pipeline `build` step, which triggers builds of other jobs.
 
+### Passing secret values to other jobs
+
+The recommended approach to pass secret values using the `build` step is to use credentials parameters:
+
+```
+build(job: 'foo', parameters: [credentials('parameter-name', 'credentials-id')])
+```
+
+See [the user guide for the Credentials Plugin](https://plugins.jenkins.io/credentials/) for a general overview of how credentials work in Jenkins and how they can be configured, and [the documentation for the Credentials Binding Plugin](https://plugins.jenkins.io/credentials-binding/) for an overview of how to access and use credentials from a Pipeline.
+
+The `build` step also supports passing password parameters, but this is not recommended.
+The plaintext secret may be persisted as part of the Pipeline's internal state, and it will not be automatically masked if it appears in the build log.
+Here is an example for reference:
+
+```
+build(job: 'foo', parameters: [password('parameter-name', 'secret-value')])
+```
+
 ## Version History
 
 See [the changelog](CHANGELOG.md).
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStep.java
@@ -157,6 +157,24 @@ public class BuildTriggerStep extends AbstractStepImpl {
             );
         }
 
+        @Override
+        public UninstantiatedDescribable customUninstantiate(UninstantiatedDescribable step) {
+            Map<String, Object> newStepArgs = copyMapReplacingEntry(step.getArguments(), "parameters", List.class, parameters -> parameters.stream()
+                    .map(parameter -> {
+                        if (parameter instanceof UninstantiatedDescribable) {
+                            UninstantiatedDescribable ud = (UninstantiatedDescribable) parameter;
+                            if (ud.getSymbol().equals("password")) {
+                                Map<String, Object> newParamArgs = copyMapReplacingEntry(ud.getArguments(), "value", Secret.class, Secret::getPlainText);
+                                return ud.withArguments(newParamArgs);
+                            }
+                        }
+                        return parameter;
+                    })
+                    .collect(Collectors.toList())
+            );
+            return step.withArguments(newStepArgs);
+        }
+
         /**
          * Copy a map, replacing the entry with the specified key if it matches the specified type.
          */
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStep.java
@@ -10,14 +10,22 @@ import hudson.model.Job;
 import hudson.model.ParameterDefinition;
 import hudson.model.ParameterValue;
 import hudson.model.ParametersDefinitionProperty;
+import hudson.model.PasswordParameterValue;
 import hudson.model.Queue;
 import hudson.util.FormValidation;
+import hudson.util.Secret;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
 import jenkins.model.Jenkins;
 import net.sf.json.JSONArray;
 import net.sf.json.JSONObject;
 import org.apache.commons.lang.StringUtils;
+import org.jenkinsci.plugins.structs.describable.CustomDescribableModel;
+import org.jenkinsci.plugins.structs.describable.DescribableModel;
+import org.jenkinsci.plugins.structs.describable.UninstantiatedDescribable;
 import org.jenkinsci.plugins.workflow.steps.AbstractStepDescriptorImpl;
 import org.jenkinsci.plugins.workflow.steps.AbstractStepImpl;
 import org.jenkinsci.plugins.workflow.steps.Step;
@@ -80,7 +88,7 @@ public class BuildTriggerStep extends AbstractStepImpl {
     }
 
     @Extension
-    public static class DescriptorImpl extends AbstractStepDescriptorImpl {
+    public static class DescriptorImpl extends AbstractStepDescriptorImpl implements CustomDescribableModel {
 
         public DescriptorImpl() {
             super(BuildTriggerStepExecution.class);
@@ -88,7 +96,7 @@ public class BuildTriggerStep extends AbstractStepImpl {
 
         // Note: This is necessary because the JSON format of the parameters produced by config.jelly when
         // using the snippet generator does not match what would be neccessary for databinding to work automatically.
-        // For non-snippet generator use, this is unnecessary.
+        // Only called via the snippet generator.
         @Override public Step newInstance(StaplerRequest req, JSONObject formData) throws FormException {
             BuildTriggerStep step = (BuildTriggerStep) super.newInstance(req, formData);
             // Cf. ParametersDefinitionProperty._doBuild:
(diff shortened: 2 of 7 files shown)


src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStep.java
```diff
@@ -122,6 +130,50 @@ public DescriptorImpl() {
...
+         * Ideally, password parameters would not be used at all with this step, but there was no documentation or
+         * runtime warnings for this usage previously and so it is relatively common.
+         */
+        @Override
+        public Map<String, Object> customInstantiate(Map<String, Object> map) {
+            if (DescribableModel.of(PasswordParameterValue.class).getParameter("value").getErasedType() != Secret.class) {
+                return map;
+            }
+            return copyMapReplacingEntry(map, "parameters", List.class, parameters -> {
+                List<Object> newParameters = new ArrayList<>(parameters.size());
```
jglick, line 144: `Stream.map` would arguably be more legible here.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepTest.java
```diff
@@ -605,7 +610,12 @@ public void invalidChoiceParameterValue() throws Exception {
 
     @Test
     public void buildStepDocs() throws Exception {
-        SnippetizerTester.assertDocGeneration(BuildTriggerStep.class);
+        try {
+            SnippetizerTester.assertDocGeneration(BuildTriggerStep.class);
+        } catch (Exception e) {
+            // TODO: Jenkins 2.236+ broke structs-based databinding and introspection of PasswordParameterValue, JENKINS-62305.
+            assumeThat(e.getMessage(), not(containsString("There's no @DataBoundConstructor on any constructor of class hudson.util.Secret")));
```
jglick, line 617: Assume that it does _not_ contain that phrase? So some unrelated NPE will be silently ignored?!

README.md
```diff
@@ -8,6 +8,24 @@
 
 Adds the Pipeline `build` step, which triggers builds of other jobs.
 
+### Passing secret values to other jobs
```
jglick, line 11: Mention also in `help-parameters.html`.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepTest.java
dwnusbaum: Yes good catch, I meant to always rethrow the exception so that it would only be skipped if the text matches and otherwise the test will fail.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStep.java
dwnusbaum: Yeah, I tried a few approaches and did not really like any of them, happy to switch to streams though.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepTest.java
dwnusbaum: Fixed in afc8024.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepTest.java
```diff
@@ -615,6 +615,7 @@ public void buildStepDocs() throws Exception {
         } catch (Exception e) {
             // TODO: Jenkins 2.236+ broke structs-based databinding and introspection of PasswordParameterValue, JENKINS-62305.
```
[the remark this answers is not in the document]
jglick: I think this _could_ be solved with `customUninstantiate` if desired.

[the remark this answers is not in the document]
jglick: Never mind, reread PR desc.

dwnusbaum:
FWIW, I realized that snippet generation for password parameters is already broken even on older versions of Jenkins. You can fix it using `customUninstantiate`, but even if you fix the output [this warning](https://github.com/jenkinsci/structs-plugin/blob/77a382758cceffdbd97c2d46f24f5d6ec8841c66/plugin/src/main/java/org/jenkinsci/plugins/structs/describable/DescribableModel.java#L674) is logged, essentially because only `BuildTriggerStep` implements `CustomDescribableModel`,  not `PasswordParameterValue` itself:

```
java.lang.ClassCastException: hudson.model.PasswordParameterValue.value expects class java.lang.String but received class hudson.util.Secret
	at org.jenkinsci.plugins.structs.describable.DescribableModel.coerce(DescribableModel.java:492)
	at org.jenkinsci.plugins.structs.describable.DescribableModel.buildArguments(DescribableModel.java:409)
	at org.jenkinsci.plugins.structs.describable.DescribableModel.instantiate(DescribableModel.java:329)
Caused: java.lang.IllegalArgumentException: Could not instantiate {name=param-name, value=secret} for hudson.model.PasswordParameterValue
	at org.jenkinsci.plugins.structs.describable.DescribableModel.instantiate(DescribableModel.java:334)
	at org.jenkinsci.plugins.structs.describable.DescribableModel.uninstantiate2(DescribableModel.java:672)
	at org.jenkinsci.plugins.structs.describable.DescribableModel.uninstantiate2_(DescribableModel.java:752)
	at org.jenkinsci.plugins.structs.describable.DescribableParameter.uncoerce(DescribableParameter.java:196)
	at org.jenkinsci.plugins.structs.describable.DescribableParameter.uncoerce(DescribableParameter.java:190)
	at org.jenkinsci.plugins.structs.describable.DescribableParameter.inspect(DescribableParameter.java:142)
	at org.jenkinsci.plugins.structs.describable.DescribableModel.uninstantiate2(DescribableModel.java:655)
	at org.jenkinsci.plugins.structs.describable.DescribableModel.uninstantiate2_(DescribableModel.java:752)
	at org.jenkinsci.plugins.workflow.steps.StepDescriptor.uninstantiate(StepDescriptor.java:252)
	at org.jenkinsci.plugins.workflow.cps.Snippetizer.object2Groovy(Snippetizer.java:139)
	at org.jenkinsci.plugins.workflow.cps.Snippetizer.step2Groovy(Snippetizer.java:81)
	at org.jenkinsci.plugins.workflow.cps.SnippetizerTester.assertRoundTrip(SnippetizerTester.java:91)
	at org.jenkinsci.plugins.workflow.support.steps.build.BuildTriggerStepTest.snippetizerRoundTrip(BuildTriggerStepTest.java:588)
	...
```

Even so, at least snippet generation works with that patch, so I'll push it to the PR branch.

src/main/resources/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStep/config.jelly
```diff
@@ -38,7 +38,7 @@ THE SOFTWARE.
     <f:entry field="quietPeriod" title="Quiet period">
         <f:number clazz="number"/>
     </f:entry>
-    <f:entry title="Parameters">
+    <f:entry title="Parameters" help="${descriptor.getHelpFile('parameters')}">
```
jglick, line 41: Does
```suggestion
    <f:entry title="Parameters" field="parameters">
```

suffice?

src/main/resources/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStep/help-parameters.html
```diff
@@ -0,0 +1,6 @@
+<div>
+    A list of parameters to pass to the downstream job.
+
```
jglick, line 3:
```suggestion
    <br>
```

?

src/main/resources/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStep/help-parameters.html
```diff
@@ -0,0 +1,6 @@
...
+    When passing secrets to downstream jobs, prefer credentials parameters over password parameters.
+    See <a href="https://plugins.jenkins.io/pipeline-build-step/">the documentation</a> for details.
```
jglick, line 5: We do not generally embed URLs into production code without redirects. Maybe invert this, keeping the details here, and having the README link to this source file or https://www.jenkins.io/doc/pipeline/steps/pipeline-build-step/#build-build-a-job ?

src/main/resources/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStep/config.jelly
dwnusbaum: Yeah, it looks like that works, added in f4d6f9b. I assumed `field` wasn't being used intentionally for some reason, but I tested and the snippet generator still seems to work correctly.

src/main/resources/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStep/help-parameters.html
dwnusbaum: I removed the blank line in c6631c5.

src/main/resources/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStep/help-parameters.html
dwnusbaum: I really prefer having docs be Markdown or AsciiDoc files in GitHub rather than linking to HTML source files or the steps reference page.

I looked around and core and quite a few plugins link to plugins.jenkins.io specifically, and the first bullet point in https://github.com/jenkinsci/jenkins/pull/2756 makes me think this is probably ok, so I am inclined to leave things as-is.

PR #48, opened 2021-02-27
Recommend the Pipeline Snippet Generator in help

MarkEWaite:
## Recommend the pipeline editor in build step help

A frequent complaint in the Pipeline Steps reference [doc comments](https://docs.google.com/spreadsheets/d/1nA8xVOkyKmZ8oTYSLdwjborT0w-BpBNNZT0nxR9deZ8/edit#gid=1087292709) is that the `wait` argument is not documented on https://www.jenkins.io/doc/pipeline/steps/pipeline-build-step/ .  The `wait` argument can't be documented on that static page because the arguments that it accepts are dynamically determined based on the installed plugins.

Suggest that the user read the online help so that they can see the best description of the values of `wait` on their Jenkins installation.


MarkEWaite:
Superseded by #49

PR #51, opened 2021-04-10
[JENKINS-65338]- Add snippet generator hint to build step help

Onyimatics:
- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your master branch!

- Added snippet generator hint to build step help

- [JENKINS-65338](https://issues.jenkins.io/browse/JENKINS-65338)

Hello @MarkEWaite @oleg-nenashev @kwhetstone @StackScribe, please kindly review.

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStep/help.html
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStep/help.html
@@ -1,3 +1,8 @@
 <div>
-    Triggers a new build for a given job.
+    <p>
+        Triggers a new build for a given job.
+    </p>
+    <p>
+        Use the <a href="https://www.jenkins.io/redirect/pipeline-snippet-generator">Pipeline Snippet Generator</a> to generate a sample pipeline script for the build step.
+    </p>
 </div>
--- a/src/main/resources/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStep/help.html
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStep/help.html
@@ -1,6 +1,6 @@
 <div>
     <p>
-        Triggers a new build for a given job.
+        Trigger a new build for a given job.
     </p>
     <p>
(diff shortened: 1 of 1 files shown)


src/main/resources/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStep/help.html
```diff
@@ -1,3 +1,8 @@
 <div>
-    Triggers a new build for a given job.
+    <p>
+        Triggers a new build for a given job.
+    </p>
+    <p>
+        Use the <a href="https://www.jenkins.io/redirect/pipeline-snippet-generator">Pipeline Snippet Generator</a> to generate a sample pipeline script for the build step.
```
oleg-nenashev, line 6: Would be better to link it to the snippet generator itself instead of its docs

MarkEWaite: Will it work to link to the snippet generator?  This page is included in both the online help inside Jenkins and the [Pipeline steps reference](https://www.jenkins.io/doc/pipeline/steps/pipeline-build-step/).  I assumed that a link to the snippet generator inside Jenkins would result in broken link complaints from readers of the Pipeline steps reference.

oleg-nenashev:
Then we should probably implement conditional logic which will show
different links depending on the target environment. It could be done via
environment variable or system property, and it could be a great
enhancement for other docs

oleg-nenashev (approved):
No objections w.r.t merging as is. Too late to address the comment about the Snippet generator. For the future, it would be nice to timely get a response from the PR author

src/main/resources/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStep/help.html
```diff
@@ -1,3 +1,8 @@
 <div>
-    Triggers a new build for a given job.
+    <p>
+        Triggers a new build for a given job.
```
MRamonLeon, line 3: Which is the best form here? I would suggest infinitive, as next paragraph.

```suggestion
        Trigger a new build for a given job.
```

Onyimatics: Thanks for the review @MRamonLeon

MRamonLeon (approved):

PR #57, opened 2021-06-25
Add hyperlink to failing downstream build

basil:
When a downstream build fails, it's helpful to have a link to click on to navigate to that failed build to investigate.


olamy (approved):

daniel-beck:
What does this do, given that https://github.com/jenkinsci/pipeline-build-step-plugin/blob/master/src/main/resources/org/jenkinsci/plugins/workflow/support/steps/build/DownstreamFailureCause/summary.jelly shows a link already?

jglick:
This was not correct. See https://github.com/jenkinsci/jenkins/pull/5653#pullrequestreview-722584558

PR #85, opened 2022-05-19
Include Pipeline features and examples to plugin documentation

Sophia-Okito:
- [x ] Please describe what you did: Include pipeline help features with descriptions and examples


README.md
```diff
@@ -6,10 +6,114 @@
...
+* upstreamBuilds: a list of upstream builds. These are the builds of the upstream projects whose artifacts feed into this build.
+* rawBuild: a hudson.model.Run with further APIs, only for trusted libraries or administrator-approved scripts outside the sandbox; the value will not be Serializable, so you may only access it inside a method marked @NonCPS.
+* keepLog: true if the log file for this build should be kept and not deleted.
+  If you do not wait, this step succeeds so long as the downstream build can be added to the queue (it will not even have been started). In that case there is currently no return value.
+
+##### Propagate error
+If enabled (default state), then the result of this step is that of the downstream build (e.g., success, unstable, failure, not built, or aborted) i.e. if the downstream build fails then the upstream build fails also. If disabled, then this step succeeds even if the downstream build is unstable, failed, etc.; use the result property of the return value as needed i.e if the downstream build fails it does not affect that of the upstream build. Example;
```
kmartens27, line 92:
```suggestion
If this option is enabled, this step inherits the status of the downstream build, for example:  success, unstable, failure, not built, or aborted.
If you disable this option, this step succeeds regardless of issues with the downstream build.
Use the result property of the return value as needed. Example;
```

kmartens27: great stuff here as well, just wanted to clean up the text a bit and make it more focused

kmartens27 (commented):
awesome stuff, let me know if there are any questions!

README.md
Sophia-Okito: Thank you, @kmartens27. All changes made

README.md
```diff
@@ -6,11 +6,117 @@
 
 ## Introduction
 
-Adds the Pipeline `build` step, which triggers builds of other jobs.
+This plugin helps to add the pipeline `build` step, which triggers builds of other jobs.
```
jglick, line 9:
```suggestion
This plugin adds the Pipeline `build` step, which triggers builds of other jobs.
```

README.md
```diff
@@ -6,11 +6,117 @@
...
+
+### Project to Build
+This is the name or path of a downstream job to build. It may be another Pipeline job, but more commonly a freestyle or other project. Use a simple name if the job is in the same folder as this upstream Pipeline job; otherwise use relative paths like `../sister-folder/downstream` or absolute paths like `/top-level-folder/nested-folder/downstream`. Example;
+```groovy
+  job: 'Hello-World'
+```
+
+### Wait for completion
+You may ask that this Pipeline build wait for completion of the downstream build. In this case, the return value of the step is an object on which you can obtain read-only properties; so you can inspect its properties such as result using `.result` and so on.
+Properties of build upon completion are;
```
jglick, line 25: Giving a few examples of how to use commonly needed properties like `number` or `result` in the context of the `build` step is great, but it is not really maintainable to enumerate all of these here. They are implemented in another plugin, to begin with, and code and docs would soon fall out of synch.

Better to point to mechanical docs. https://www.jenkins.io/doc/pipeline/steps/ does not seem to have it, but live inside Jenkins you can go to e.g. https://ci.jenkins.io/pipeline-syntax/globals#currentBuild

adampl:
Please also add info how to pass normal parameters.

jglick:
Seems to be abandoned.

PR #105, opened 2023-02-14
Implement option to wait only until the downstream build is started

stuartrowe:
Add a new option `waitForStart` to the `build` step. The `waitForStart` option changes the behavior of the `build` step to block until the downstream build is started and a valid `RunWrapper` can be returned.

Implements https://issues.jenkins.io/browse/JENKINS-60849

- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [x] Link to relevant issues in GitHub or Jira
- [x] Link to relevant pull requests, esp. upstream and downstream changes
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue

PR #106, opened 2023-02-14
Implement waitForBuild step

stuartrowe:
Add a new `waitForBuild` pipeline step that waits for a given build to complete. The build to wait on is defined by its [`externalizableId`](https://github.com/jenkinsci/jenkins/blob/62c2fbcc7af34a2bc4edd67e46939fbe90f3b899/core/src/main/java/hudson/model/Run.java#L2467).

Implements https://issues.jenkins.io/browse/JENKINS-70617.

This PR is intended to be paired with https://github.com/jenkinsci/pipeline-build-step-plugin/pull/105. The source branch is currently based off of https://github.com/stuartrowe/pipeline-build-step-plugin/tree/JENKINS-60849.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/WaitForBuildAction.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/WaitForBuildAction.java
@@ -1,0 +1,15 @@
+package org.jenkinsci.plugins.workflow.support.steps.build;
+
+import hudson.model.InvisibleAction;
+import org.jenkinsci.plugins.workflow.steps.StepContext;
+
+public class WaitForBuildAction  extends InvisibleAction {
+
+    final StepContext context;
+    final boolean propagate;
+
+    WaitForBuildAction(StepContext context, boolean propagate) {
+        this.context = context;
+        this.propagate = propagate;
+    }
+}
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/WaitForBuildListener.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/WaitForBuildListener.java
@@ -1,0 +1,33 @@
+package org.jenkinsci.plugins.workflow.support.steps.build;
+
+import edu.umd.cs.findbugs.annotations.NonNull;
+import hudson.Extension;
+import hudson.model.Result;
+import hudson.model.Run;
+import hudson.model.TaskListener;
+import hudson.model.listeners.RunListener;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import org.jenkinsci.plugins.workflow.steps.FlowInterruptedException;
+import org.jenkinsci.plugins.workflow.steps.StepContext;
+
+@Extension
+public class WaitForBuildListener extends RunListener<Run<?,?>> {
+
+    private static final Logger LOGGER = Logger.getLogger(WaitForBuildListener.class.getName());
+
+    @Override
+    public void onCompleted(Run<?,?> run, @NonNull TaskListener listener) {
+        for (WaitForBuildAction action : run.getActions(WaitForBuildAction.class)) {
+            StepContext context = action.context;
+            LOGGER.log(Level.FINE, "completing {0} for {1}", new Object[] {run, context});
+            if (!action.propagate || run.getResult() == Result.SUCCESS) {
+                context.onSuccess(new RunWrapper(run, false));
+            } else {
+                Result result = run.getResult();
+                context.onFailure(new FlowInterruptedException(result != null ? result : /* probably impossible */ Result.FAILURE, false, new DownstreamFailureCause(run)));
+            }
+        }
+        run.removeActions(WaitForBuildAction.class);
+    }
+}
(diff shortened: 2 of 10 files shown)


stuartrowe:
@jglick @dwnusbaum would either of you be able to review this or suggest someone else? I'm not sure who's maintaining this currently.

olamy:
we should probably wait for https://github.com/jenkinsci/pipeline-build-step-plugin/pull/105  to be merged first?

stuartrowe:
Yes, I'll rebase once https://github.com/jenkinsci/pipeline-build-step-plugin/pull/105 is merged.

olamy (approved):

olamy:
> @olamy is there anything preventing this from being merged?

a ping was missing ;)

olamy:
@stuartrowe sorry for delay not checking this. Merge it now

hamaney:
Good stuff. I always wanted this feature. Cheers.

r00t-:
this is extremely useful, thanks.

i think it should come with an example of how to use it.
specifically, it's confusing how `waitForBuild()` takes an `externalizableId`. yet it seems there is no public api to obtain that from a build started from a pipeline script.
i found `build(...,waitForStart: true).getRawBuild().getExternalizableId()`, but that requires to add pipeline security exceptions for those functions.

stuartrowe:
@r00t- that's a fair point. It would be nice if [RunWrapper](https://github.com/jenkinsci/workflow-support-plugin/blob/master/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java) exposed `getExternalizableId()` in its API. I can certainly submit a PR for that.

Your example works, but I wouldn't advise using it directly from pipeline code -  `getRawBuild()` returns a [Run](https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/model/Run.java) which isn't serializable and shouldn't be referenced in CPS transformed code. 

We have a `@NonCPS` helper function in our shared library for extracting the `externalizableId`:

```
@NonCPS
static String getExternalizableId(RunWrapper runWrapper) {
	def externalizableId = null
	if (runWrapper) {
		def run = runWrapper.getRawBuild()
		if (run) {
			externalizableId = run.getExternalizableId()
		}
	}
	return externalizableId
}
```

I'm not sure if that's the best candidate for example code though because it assumes the use of a scripted vs declarative pipeline. I'd like to try and get `RunWrapper#getExternalizableId()` added before updating the documentation.

jglick:
The externalizable id is simply the job full name plus `#` plus the build number.

stuartrowe:
Yes, but that's a bit clunky in pipeline code. Since [Run#getExternalizableId()](https://github.com/jenkinsci/jenkins/blob/57569071c7c093890c424356257536d2fd11f1ac/core/src/main/java/hudson/model/Run.java#L2460) already exists it would be convenient to expose it for `RunWrapper`  too.

stuartrowe:
Very simple PR opened here: https://github.com/jenkinsci/workflow-support-plugin/pull/268

r00t-:
@stuartrowe:
i just removed our production use of `waitForBuild()`, because we found that it breaks `parallel([failfast: true, ...])`,
which lead to loads of wasted resources on build that should have been aborted.

any idea how this happens and if it can be fixed?

here's a simple repro:
```
parallel([
    failFast: true,
    fail: {
        def b=build(job: "slow job", waitForStart: true)
        def runid = b.getRawBuild().getExternalizableId()
        waitForBuild(runId: runid, propagate: false)
    },
    slow: {
        def b=build(job: "fail job", waitForStart: true)
        def runid = b.getRawBuild().getExternalizableId()
        waitForBuild(runId: runid, propagate: false)
    },
])
```
(where `fail job` is a job that fails immediately (`sh "exit 1"), and `slow job` is a job that simply sleeps for a minute (`sh "sleep 240"`)
the alternative of
```
build([job: ..., wait: true, propagate: false])
```
does not have that issue.

stuartrowe:
Have you tried adding `propagateAbort: true` to the `waitForBuild` step? I haven't looked into the mechanism for `parallel`'s failfast but I'm guessing it uses abort to fail the other parallel branches.

r00t-:
thanks for the quick response!
```
        waitForBuild(runId: runid, propagate: false, propagateAbort: true)
```
doesn't change the behavior

r00t-:
from my own research, failfast normally causes
the builds to catch a `FlowInterruptedException` with a cause of `BuildTriggerCancelledCause`: "calling pipeline cancelled"

stuartrowe:
Alright, that's unfortunate. I'll have to look into it then.

r00t-:
thanks so far!

r00t-:
my experience with jenkins' internal workings are limited...
but this should be the code that triggers the cancellation for failfast.
i have no way to tell how it could break (or how it works) :sweat_smile: .
```
handler.stepExecution.stop(new FlowInterruptedException(Result.ABORTED, true, new FailFastCause(name)));
```
https://github.com/jenkinsci/workflow-cps-plugin/blob/dc9d3aefcc6a28e8b5efa430a23d5b04ae0d92d1/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ParallelStep.java#L151

jglick:
Offhand I would suspect the wrong value of https://javadoc.jenkins.io/plugin/workflow-step-api/org/jenkinsci/plugins/workflow/steps/FlowInterruptedException.html#isActualInterruption() somewhere, but in general it is best to reproduce any claimed issue in a `JenkinsRule` test before studying root causes and potential fixes.

PR #116, opened 2023-05-24
[JENKINS-71342] WaitForBuild should proagate result of already completed downstream builds when propagate is enabled.

stuartrowe:
https://issues.jenkins.io/browse/JENKINS-71342

Handle the result of the downstream build being waited on by the `waitForBuild` step when it has already completed and `propagate` is enabled. In this case the `waitForBuild` step will call `getContext().onFailure` with a `FlowInterruptedException` wrapping the result of the downstream run along with the underlying `DownstreamFailureCause`.

### Testing done

Added a new unit test `WaitForBuildStepTest#waitForBuildPropagateAlreadyCompleteFailure` to confirm the issue and validate the fix.

Fixes: https://issues.jenkins.io/browse/JENKINS-71342.

### Submitter checklist


jglick:
> Do you think it's still necessary to address the test reliability?

Probably not—it should now pass either way. Still not ideal that the test is usually testing one code path, but occasionally testing another.

stuartrowe:
@jglick I've also updated the tests to address the issues - could either you or @olamy take a look?

stuartrowe:
@jglick @olamy poking again for a review.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/WaitForBuildStepExecution.java
```diff
@@ -34,8 +36,18 @@ public boolean start() throws Exception {
             taskListener.getLogger().println("Waiting for " + runHyperLink + " to complete");
             return false;
         } else {
-            taskListener.getLogger().println(runHyperLink + " is already complete");
-            getContext().onSuccess(null);
+            Result result = run.getResult();
+            if (result == null) {
```
jtnord, line 40: this should never occur (for a [pipeline job](https://github.com/jenkinsci/workflow-job-plugin/blob/d2290d3341af6011b1f7833ab57d4db28b824e3a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java#L797-L799C6) or a [FreeStyle](https://github.com/jenkinsci/jenkins/blob/019950a4a302db7f1094d49be9025866307356d5/core/src/main/java/hudson/model/Run.java#L527-L534) job due to its [state](https://github.com/jenkinsci/jenkins/blob/019950a4a302db7f1094d49be9025866307356d5/core/src/main/java/hudson/model/Run.java#L266-L275),   as such I would this should at least be shown in the logs that the build is in an indeterminate state

```suggestion
            if (result == null) {
                taskListener.getLogger().println("Warning: " + runHyperLink + " already completed but getResult() returned null.  Treating the result of this build as a failure));
```

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/WaitForBuildStepExecution.java
```diff
@@ -34,8 +36,18 @@ public boolean start() throws Exception {
...
+                result = Result.FAILURE;
+            }
+            taskListener.getLogger().println(runHyperLink + " already completed: " + result.toString());
```
jtnord, line 43:
```suggestion
            else  {
                taskListener.getLogger().println(runHyperLink + " already completed: " + result.toString());
            }
```

stuartrowe:
@jtnord could you please take another look at this when you have a chance?

jtnord (approved):

jtnord:
thanks for the fix @stuartrowe

PR #124, opened 2023-09-22
[JENKINS-71961] Fix waitForBuild Step does not abort downstream build when upstream job aborted

stuartrowe:
Implemented `stop()` method in `WaitForBuildStepExecution` so aborts can be propagated from the upstream job to downstream job that is being waited on.

Fix for https://issues.jenkins.io/browse/JENKINS-71961.

### Testing done

Add two new test cases to `WaitForBuildStepTest`:
- `abortBuild()`: validates that when the downstream job being waited on is aborted that the upstream job will be aborted too.
- `interruptFlow()`: validates that when the upstream job aborted then downstream job being waited on will be aborted too.

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/WaitForBuildStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/WaitForBuildStepExecution.java
@@ -3,16 +3,26 @@ package org.jenkinsci.plugins.workflow.support.steps.build;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import hudson.AbortException;
 import hudson.console.ModelHyperlinkNote;
+import hudson.model.Computer;
+import hudson.model.Executor;
+import hudson.model.Queue;
 import hudson.model.Result;
 import hudson.model.Run;
 import hudson.model.TaskListener;
+import jenkins.model.Jenkins;
 import org.jenkinsci.plugins.workflow.steps.AbstractStepExecutionImpl;
 import org.jenkinsci.plugins.workflow.steps.FlowInterruptedException;
 import org.jenkinsci.plugins.workflow.steps.StepContext;
 
+import java.io.IOException;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
 public class WaitForBuildStepExecution extends AbstractStepExecutionImpl {
 
     private static final long serialVersionUID = 1L;
+
+    private static final Logger LOGGER = Logger.getLogger(WaitForBuildStepExecution.class.getName());
 
     private final transient WaitForBuildStep step;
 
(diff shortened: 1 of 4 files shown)


stuartrowe:
Closing to re-run tests.

stuartrowe:
Reopening to re-run tests.

stuartrowe:
@jenkinsci/pipeline-build-step-plugin-developers there's [a discussion](https://issues.jenkins.io/browse/JENKINS-71961?focusedId=441640&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-441640) about the behavior of this feature.

The reporter of the issue and I both agree that the downstream run shouldn't always be aborted when the waiting build is aborted. 

I am hoping for some input on whether adding a parameter such as `propagateAbort` to the `waitForBuild` step would be appropriate. If that approach is reasonable then is there a better name to avoid confusion with the existing `propagate` parameter?

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/WaitForBuildStepExecution.java
```diff
@@ -55,4 +65,51 @@ public boolean start() throws Exception {
...
+        boolean interrupted = false;
+
+        // if there's any in-progress build already, abort that.
+        // when the build is actually aborted, WaitForBuildListener will take notice and report the failure,
+        // so this method shouldn't call getContext().onFailure()
+        for (Computer c : jenkins.getComputers()) {
+            for (Executor e : c.getExecutors()) {
+                interrupted |= maybeInterrupt(e, cause, context);
+            }
+            for (Executor e : c.getOneOffExecutors()) {
```
jglick, line 86:
```suggestion
            for (Executor e : c.getAllExecutors()) {
```

src/test/java/org/jenkinsci/plugins/workflow/support/steps/build/WaitForBuildStepTest.java
```diff
@@ -112,7 +178,7 @@ private static FlowNode findFirstNodeWithDescriptor(FlowExecution execution, Cla
         return null;
     }
 
-    private WorkflowJob createWaitingDownStreamJob(Result result) throws Exception {
+    private WorkflowJob createWaitingDownStreamJob(Result result, boolean indefiniteWait) throws Exception {
         WorkflowJob ds = j.jenkins.createProject(WorkflowJob.class, "ds");
         ds.setDefinition(new CpsFlowDefinition(
             "import org.jenkinsci.plugins.workflow.support.steps.build.WaitForBuildAction\n" +
```
jglick, line 184: BTW it would be more natural to use `semaphore` and then have test code unblock once some condition is satisfied.

jglick (commented):
Seems reasonable from a quick glance (I have not followed development of this step in detail). No particular opinion offhand about making this behavior an option, or what the default value should be.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/build/WaitForBuildStepTest.java
stuartrowe: Agreed, but I can't think of a way to coordinate that between the test code and the pipeline implementation of the test jobs.

stuartrowe:
I've updated the behavior so that aborting an upstream build currently waiting on a downstream build will only abort that downstream build if it was was originally triggered by the upstream build. The rationale is so these two example pipelines will have the same behavior:
```
build(job: "downstream") // aborted while this step is still waiting for the 'downstream' build to complete
```

```
def ds = build(job: "downstream", waitForStart: true)
// do something with the in-progress 'downstream' build
waitForBuild(runId: "${ds.getFullProjectName()}#${ds.getId()}") // aborted while this step is still waiting for the 'downstream' build to complete
```

stuartrowe:
Re-running tests.

stuartrowe:
I ended up going back to the solution using an additional `propagateAbort` argument. This is working well on my local instance.

@jglick please review if you have time. Thanks!

matthuisman:
yes please :)

stuartrowe:
@matthuisman if you call `build` with `wait: true` then an abort will also abort the triggered downstream build. I don't think it's practical to pass `propagateAbort` to the `build` call and then have that take effect while waiting on a subsequent `waitForBuild` call though.

stuartrowe:
@car-roll are you able to review this change?

stuartrowe:
@jglick @car-roll @dwnusbaum bumping for review.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/build/WaitForBuildStepTest.java
```diff
@@ -112,22 +252,23 @@ private static FlowNode findFirstNodeWithDescriptor(FlowExecution execution, Cla
...
         return ds;
-        
+    }
+
+    private void waitForWaitForBuildAction(WorkflowRun r) throws Exception {
+        while(true) {
+            if (r.getAction(WaitForBuildAction.class) != null) {
+                break;
+            }
+            Thread.sleep(10);
```
jglick, line 270: BTW I recommend Awaitility.

jglick (approved):

matthuisman:
@stuartrowe 
Can you show example of using propagateAbort ?

matthuisman:
does this only work once?

I have a main job that i abort.
I then have it loop its child jobs and do waitForBuild with propagateAbort

It aborts the current job its waiting on - but when it goes to the next running job - it doesnt abort that

stuartrowe:
@matthuisman can you provide a sample pipeline that reproduces the issue you're seeing? Only downstream jobs that are currently being waited on through `waitForBuild` would be aborted when the parent job is aborted.

PR #126, opened 2023-10-09
Remove inline javascript code from BuildTriggerStep

NeetigyaPod:
### Testing done
Inspected the element from local machine to check the change. Screenshots attached

Before:
image

After:
Screenshot from 2023-10-09 22-46-02

```[tasklist]
### Submitter checklist
```
Changed the inline script with accordance to [Security Guideline for inline scripts](https://www.jenkins.io/doc/developer/security/csp/#inline-javascript-blocks).
Issue : [Issue #72145] (https://issues.jenkins.io/browse/JENKINS-72145)


jglick:
check `BuildTriggerStepConfigTest`

NeetigyaPod:
Hi, the tests seem to fail from original code
Screenshot from 2023-10-09 23-38-00
Can you please guide me in the same?

jglick:
> the tests seem to fail from original code

https://github.com/jenkinsci/pipeline-build-step-plugin/runs/17351729476 passed in trunk last week. Also, the test failures in this PR are covering exactly the feature area you are modifying, so I think they are very likely related.

NeetigyaPod:
My bad, I had left the script tags there. Rectified it, is there any way I can check if tests pass locally? I did mvn test but they fail locally. Do I have to download any requirements to set it up locally?
Right now I cloned and am just mvn:hpi run and checking.

jglick:
> Do I have to download any requirements to set it up locally?

Just latest Java 11 and latest Maven.

src/main/resources/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStep/configLoad.js
```diff
@@ -0,0 +1,13 @@
+function loadParams() {
+    var div = document.getElementById('params');
+    fetch('${descriptor.descriptorUrl}/parameters?job=' + encodeURIComponent(document.getElementById('${jobFieldId}').value) + '&amp;context=${descriptor.contextEncoded}').then((rsp) => {
```
NeetigyaPod, line 3: Hi, I have solved the Test issues (needed to mvn clean and then test and remove script tags). I need a bit of help in how to pass jelly values to javascript. Stapler adjuncts dont allow to pass such variables and it is causing null pointer exceptions in one test case because it cannot find the HTML element

jglick: I am not the expert, but I recall that the usual pattern is to [use `data-*` attributes](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes).

NeetigyaPod: Hi, thank you for being patient. The solution worked and the tests are passing now.

NeetigyaPod:
Also can a "hacktoberfest-accepted" label be added to this PR? Will be grateful for that since the jira issue already has the tag.

NeetigyaPod:
Hi have made the changes, the bugs are fixed

Kevin-CB (changes requested):
There is also an `onblur` call on this page.

https://github.com/jenkinsci/pipeline-build-step-plugin/blob/2239584eebab2a4d7a0e8d0d7ae5ddf3e07834da/src/main/resources/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStep/config.jelly#L30

Kevin-CB (approved):
I tested it locally, it works as expected and I'm not able to see CSP violation using the CSP plugin

NeetigyaPod:
@jglick Hi, can we please add the hacktoberfest topic in this repository? Or a label of hacktoberfest-accepted to this PR works as well.

src/main/resources/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStep/config.jelly
```diff
@@ -27,7 +27,9 @@ THE SOFTWARE.
 <j:jelly xmlns:j="jelly:core" xmlns:f="/lib/form" xmlns:st="jelly:stapler">
     <j:set var="jobFieldId" value="${h.generateId()}"/>
```
jglick, line 28: In general I think we should be able to get rid of `h.generateId()` in these refactorings, by making the behavior function take the element argument into account and looking for siblings. Same for the hard-coded id `params`, which we did not even bother uniquifying using `generateId` since we expect this Jelly facet to be loaded only once in a page (**Snippet generator**).

jglick (approved):

src/main/resources/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStep/config.jelly
NeetigyaPod: Oh makes sense, I want to contribute to Jenkins in long term as well so I can work on it as a separate issue?
Also is there any slack for Jenkins or is Gitter the preferred platform?

jglick: > I can work on it as a separate issue?

Sure, if you would like. It is not a priority, would just be a nice stylistic improvement.

> is there any slack for Jenkins or is Gitter the preferred platform?

https://www.jenkins.io/chat/ though TBH I am not normally on those. Not sure why Slack is not used.

PR #127, opened 2023-12-08
Add `DownstreamBuildAction` to track metadata for scheduled downstream builds on the upstream `FlowNode`

dwnusbaum:
I am working on a Pipeline visualization and want to show special information about builds started by the build being visualized. I did not realize that there is not really any good way to get this information currently. It is technically possible to set up a `RunListener` that looks at `BuildUpstreamCause` as Blue Ocean does (see [DownstreamJobListener](https://github.com/jenkinsci/blueocean-plugin/blob/f295a17dbf15f06abc31964ac7dab08efebd1db7/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/listeners/DownstreamJobListener.java)), but it seems silly to have every visualization plugin that wants this information set up a redundant listener with the same logic.

This PR adds a `DownstreamBuildAction` on the `FlowNode` for the `build` step. At first it only has the full name of the triggered job. Once a build starts, it also holds the build number and externalizable ID for the build.

CC @artmorse @jgreffe

### Testing done

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerListener.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerListener.java
@@ -106,13 +106,14 @@ public class BuildTriggerListener extends RunListener<Run<?,?>>{
                             LOGGER.log(Level.FINE, () -> "Unable to update DownstreamBuildAction for " + upstream + " node " + buildUpstreamCause.getNodeId());
                             continue;
                         }
-                        DownstreamBuildAction downstreamAction = node.getAction(DownstreamBuildAction.class);
+                        DownstreamBuildAction downstreamAction = node.getPersistentAction(DownstreamBuildAction.class);
                         if (downstreamAction == null) {
                             // Should only happen for builds already in the queue when this plugin is updated to include DownstreamBuildAction.
                             downstreamAction = new DownstreamBuildAction(run.getParent());
                             node.addAction(downstreamAction);
                         }
                         downstreamAction.setBuild(run);
+                        run.save();
                     } catch (IOException e) {
                         LOGGER.log(Level.FINE, e, () -> "Unable to update DownstreamBuildAction for " + upstream + " node " + buildUpstreamCause.getNodeId());
                     }
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerListener.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerListener.java
@@ -3,14 +3,18 @@ package org.jenkinsci.plugins.workflow.support.steps.build;
 import hudson.AbortException;
 import hudson.Extension;
 import hudson.console.ModelHyperlinkNote;
+import hudson.model.Cause;
 import hudson.model.Result;
 import hudson.model.Run;
 import hudson.model.TaskListener;
 import hudson.model.listeners.RunListener;
+import java.io.IOException;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import jenkins.util.Timer;
 import org.jenkinsci.plugins.workflow.actions.WarningAction;
+import org.jenkinsci.plugins.workflow.flow.FlowExecution;
+import org.jenkinsci.plugins.workflow.flow.FlowExecutionOwner;
 import org.jenkinsci.plugins.workflow.graph.FlowNode;
 import org.jenkinsci.plugins.workflow.steps.FlowInterruptedException;
 import org.jenkinsci.plugins.workflow.steps.StepContext;
@@ -40,6 +44,7 @@ public class BuildTriggerListener extends RunListener<Run<?,?>>{
                 LOGGER.log(Level.FINE, "{0} unavailable in {1}", new Object[] {stepContext, run});
             }
         }
+        Timer.get().submit(() -> updateDownstreamBuildAction(run));
     }
 
     @Override
@@ -80,6 +85,39 @@ public class BuildTriggerListener extends RunListener<Run<?,?>>{
     public void onDeleted(final Run<?,?> run) {
         for (final BuildTriggerAction.Trigger trigger : BuildTriggerAction.triggersFor(run)) {
             Timer.get().submit(() -> trigger.context.onFailure(new AbortException(run.getFullDisplayName() + " was deleted")));
+        }
+    }
+
+    private void updateDownstreamBuildAction(Run<?, ?> run) {
+        for (Cause cause : run.getCauses()) {
+            if (cause instanceof BuildUpstreamCause) {
+                BuildUpstreamCause buildUpstreamCause = (BuildUpstreamCause) cause;
+                Run<?, ?> upstream = buildUpstreamCause.getUpstreamRun();
+                if (upstream instanceof FlowExecutionOwner.Executable) {
+                    FlowExecutionOwner owner = ((FlowExecutionOwner.Executable) upstream).asFlowExecutionOwner();
+                    if (owner == null) {
+                        LOGGER.log(Level.FINE, () -> "Unable to update DownstreamBuildAction for " + upstream + " node " + buildUpstreamCause.getNodeId());
+                        continue;
+                    }
+                    try {
+                        FlowExecution execution = owner.get();
+                        FlowNode node = execution.getNode(buildUpstreamCause.getNodeId());
+                        if (node == null) {
+                            LOGGER.log(Level.FINE, () -> "Unable to update DownstreamBuildAction for " + upstream + " node " + buildUpstreamCause.getNodeId());
+                            continue;
+                        }
+                        DownstreamBuildAction downstreamAction = node.getAction(DownstreamBuildAction.class);
+                        if (downstreamAction == null) {
+                            // Should only happen for builds already in the queue when this plugin is updated to include DownstreamBuildAction.
+                            downstreamAction = new DownstreamBuildAction(run.getParent());
+                            node.addAction(downstreamAction);
+                        }
+                        downstreamAction.setBuild(run);
+                    } catch (IOException e) {
+                        LOGGER.log(Level.FINE, e, () -> "Unable to update DownstreamBuildAction for " + upstream + " node " + buildUpstreamCause.getNodeId());
+                    }
+                }
+            }
         }
     }
 }
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildUpstreamNodeAction.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildUpstreamNodeAction.java
@@ -1,6 +1,5 @@
 package org.jenkinsci.plugins.workflow.support.steps.build;
 
-import edu.umd.cs.findbugs.annotations.CheckForNull;
 import hudson.model.InvisibleAction;
 import hudson.model.Run;
 import org.jenkinsci.plugins.workflow.graph.FlowNode;
@@ -11,7 +10,9 @@ import java.util.Objects;
  * Attached to newly-created builds in order to point back to the triggering FlowNode.
  *
  * @see DownstreamBuildAction
+ * @deprecated Use {@link BuildUpstreamCause} instead
  */
+@Deprecated
 public class BuildUpstreamNodeAction extends InvisibleAction {
 
     private final String upstreamNodeId;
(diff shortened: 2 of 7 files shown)


src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildUpstreamNodeAction.java
```diff
@@ -9,11 +10,7 @@
 /**
  * Attached to newly-created builds in order to point back to the triggering FlowNode.
  *
- * We annotate the downstream build instead of upstream in order to support no-wait builds,
- * because the downstream run id is not available when they're still queued.
- *
- * Needed for Blue Ocean to annotate the correct step.
- * See <a href="https://issues.jenkins-ci.org/browse/JENKINS-38339">#JENKINS-38339</a>
```
dwnusbaum, line 16: Blue Ocean uses `BuildUpstreamCause`, not the action (and fwiw the feature seems to be partially broken based on comments in that tickets and others).

jglick: AFAICT this is used only by https://github.com/jenkinsci/depgraph-view-plugin/blob/1046773fb05c114b876c09a0fcff7863fbb8da34/src/main/java/hudson/plugins/depgraph_view/model/graph/edge/BuildTriggerEdgeProvider.java#L81-L92 and so could be deprecated in favor of `BuildUpstreamCause` which does essentially the same thing?

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/DownstreamBuildAction.java
```diff
@@ -0,0 +1,53 @@
...
+import hudson.model.ItemGroup;
+import hudson.model.Run;
+import org.springframework.security.access.AccessDeniedException;
+
+/**
+ * Tracks the downstream build triggered by the {@code build} step.
+ * <p>Note that {@link #getBuildNumber} may return null if the build is in the queue or gets cancelled from the queue.
+ * <p>Keep in mind that {@link #getBuild} may need to load the build, and so should be used carefully.
+ *
+ * @see BuildUpstreamCause
```
jglick, line 16: Javadoc link from `BuildUpstreamCause` → `DownstreamBuildAction` would be helpful as well.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/DownstreamBuildAction.java
```diff
@@ -0,0 +1,53 @@
...
+ * @see BuildUpstreamNodeAction
+ */
+public class DownstreamBuildAction extends InvisibleAction {
```
jglick, line 19:
```suggestion
public final class DownstreamBuildAction extends InvisibleAction {
```

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/DownstreamBuildAction.java
```diff
@@ -0,0 +1,53 @@
...
+    private final String jobFullName;
+    private Integer buildNumber;
+    private String buildId;
+
+    public DownstreamBuildAction(Item job) {
```
jglick, line 24:
```suggestion
    DownstreamBuildAction(Item job) {
```

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/DownstreamBuildAction.java
```diff
@@ -0,0 +1,53 @@
...
+import edu.umd.cs.findbugs.annotations.NonNull;
+import hudson.model.InvisibleAction;
+import hudson.model.Item;
+import hudson.model.ItemGroup;
+import hudson.model.Run;
+import org.springframework.security.access.AccessDeniedException;
+
+/**
+ * Tracks the downstream build triggered by the {@code build} step.
+ * <p>Note that {@link #getBuildNumber} may return null if the build is in the queue or gets cancelled from the queue.
```
jglick, line 13: Better moved to Javadoc of that method.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/DownstreamBuildAction.java
```diff
@@ -0,0 +1,53 @@
...
+        return buildNumber;
+    }
+
+    /**
+     * Load the downstream build, if it has started and still exists.
+     * <p>Loading builds indiscriminately will affect controller performance, so use this carefully. If you only need
+     * to know whether the build started at one point, use {@link #getBuildNumber}.
+     * @throws AccessDeniedException as per {@link ItemGroup#getItem}
+     */
+    public @CheckForNull Run<?, ?> getBuild() {
```
jglick, line 42:
```suggestion
    public @CheckForNull Run<?, ?> getBuild() throws AccessDeniedException {
```

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/DownstreamBuildAction.java
```diff
@@ -0,0 +1,53 @@
...
+ * <p>Note that {@link #getBuildNumber} may return null if the build is in the queue or gets cancelled from the queue.
+ * <p>Keep in mind that {@link #getBuild} may need to load the build, and so should be used carefully.
+ *
+ * @see BuildUpstreamCause
+ * @see BuildUpstreamNodeAction
+ */
+public class DownstreamBuildAction extends InvisibleAction {
+    private final String jobFullName;
+    private Integer buildNumber;
+    private String buildId;
```
jglick, line 22: This is simply `${jobFullName}#${buildNumber}` so it does not make sense to store separately.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerListener.java
```diff
@@ -82,4 +87,37 @@ public void onDeleted(final Run<?,?> run) {
...
+                            LOGGER.log(Level.FINE, () -> "Unable to update DownstreamBuildAction for " + upstream + " node " + buildUpstreamCause.getNodeId());
+                            continue;
+                        }
+                        DownstreamBuildAction downstreamAction = node.getAction(DownstreamBuildAction.class);
+                        if (downstreamAction == null) {
+                            // Should only happen for builds already in the queue when this plugin is updated to include DownstreamBuildAction.
+                            downstreamAction = new DownstreamBuildAction(run.getParent());
+                            node.addAction(downstreamAction);
+                        }
+                        downstreamAction.setBuild(run);
```
jglick, line 115:
```suggestion
                        downstreamAction.setBuild(run);
                        run.save();
```

jglick (approved):

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/DownstreamBuildAction.java
dwnusbaum: Yeah, I prefer not to try to interpret/parse otherwise opaque IDs outside of the context where they are defined, but I guess it is already effectively impossible to change the structure of `Run.getExternalizableId` so we can just build the ID ourselves from the components.

jglick: Yes I would consider the format of the externalizable id to effectively be a committed API. There are certainly other pieces of code which assume it can be constructed using string concatenation.

dwnusbaum: https://github.com/jenkinsci/pipeline-build-step-plugin/pull/127/commits/1a96a2b25711b9f2b6cde3632747747e67faaaf4

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/DownstreamBuildAction.java
dwnusbaum: https://github.com/jenkinsci/pipeline-build-step-plugin/pull/127/commits/1a96a2b25711b9f2b6cde3632747747e67faaaf4

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/DownstreamBuildAction.java
dwnusbaum: https://github.com/jenkinsci/pipeline-build-step-plugin/pull/127/commits/1a96a2b25711b9f2b6cde3632747747e67faaaf4

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/DownstreamBuildAction.java
dwnusbaum: https://github.com/jenkinsci/pipeline-build-step-plugin/pull/127/commits/1a96a2b25711b9f2b6cde3632747747e67faaaf4

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/DownstreamBuildAction.java
dwnusbaum: https://github.com/jenkinsci/pipeline-build-step-plugin/pull/127/commits/1a96a2b25711b9f2b6cde3632747747e67faaaf4

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/DownstreamBuildAction.java
dwnusbaum: https://github.com/jenkinsci/pipeline-build-step-plugin/pull/127/commits/1a96a2b25711b9f2b6cde3632747747e67faaaf4

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildUpstreamNodeAction.java
dwnusbaum: Yes I think so, I marked it as deprecated in https://github.com/jenkinsci/pipeline-build-step-plugin/pull/127/commits/1a96a2b25711b9f2b6cde3632747747e67faaaf4.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/DownstreamBuildAction.java
```diff
@@ -6,29 +6,29 @@
...
+    DownstreamBuildAction(Item job) {
         this.jobFullName = job.getFullName();
     }
 
     public @NonNull String getJobFullName() {
         return jobFullName;
     }
 
+    /**
+     * Get the build number of the downstream build, or {@code null} if the downstream build has not yet started.
```
jglick, line 30:
```suggestion
     * Get the build number of the downstream build, or {@code null} if the downstream build has not yet started or the queue item was cancelled.
```
?

jglick (approved):

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerListener.java
```diff
@@ -82,4 +87,38 @@ public void onDeleted(final Run<?,?> run) {
...
+                            continue;
+                        }
+                        DownstreamBuildAction downstreamAction = node.getPersistentAction(DownstreamBuildAction.class);
+                        if (downstreamAction == null) {
+                            // Should only happen for builds already in the queue when this plugin is updated to include DownstreamBuildAction.
+                            downstreamAction = new DownstreamBuildAction(run.getParent());
+                            node.addAction(downstreamAction);
+                        }
+                        downstreamAction.setBuild(run);
+                        run.save();
```
Artmorse, line 116: During my tests it seems that restarting the controller reset the `DownstreamAction` (at least its `buildNumber` property).
Do we need to save the action?

jglick: `run.save()` ought to do that. If you suspect this is broken somehow, it should be possible to reproduce using `JenkinsSessionRule`: run through this scenario in one session, then restart Jenkins and recheck the metadata.

dwnusbaum: Oh, I bet the problem is that the `FlowNodeStorage` have optimizations to try to only save nodes that have changed, and setting a field on a single action by itself will not mark the `FlowNode` as updated in a way that the storages will recognize. I will look into it.

jglick: Mutable fields :facepalm: maybe better to just make fields `final` and use `replaceAction`?

dwnusbaum: Yes that was my plan

PR #128, opened 2023-12-14
Store `DownstreamBuildAction` on `Run` instead of `FlowNode` for consistent persistence even on completed builds

dwnusbaum:
See https://github.com/jenkinsci/pipeline-build-step-plugin/pull/127#discussion_r1426716551.

CC @artmorse

### Testing done

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerListener.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerListener.java
@@ -106,13 +106,8 @@ public class BuildTriggerListener extends RunListener<Run<?,?>>{
                             LOGGER.log(Level.FINE, () -> "Unable to update DownstreamBuildAction for " + upstream + " node " + buildUpstreamCause.getNodeId());
                             continue;
                         }
-                        DownstreamBuildAction downstreamAction = node.getPersistentAction(DownstreamBuildAction.class);
-                        if (downstreamAction == null) {
-                            // Should only happen for builds already in the queue when this plugin is updated to include DownstreamBuildAction.
-                            downstreamAction = new DownstreamBuildAction(run.getParent());
-                            node.addAction(downstreamAction);
-                        }
-                        downstreamAction.setBuild(run);
+                        DownstreamBuildAction action = new DownstreamBuildAction(run);
+                        node.addOrReplaceAction(action);
                         run.save();
                     } catch (IOException e) {
                         LOGGER.log(Level.FINE, e, () -> "Unable to update DownstreamBuildAction for " + upstream + " node " + buildUpstreamCause.getNodeId());
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerListener.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerListener.java
@@ -13,7 +13,6 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import jenkins.util.Timer;
 import org.jenkinsci.plugins.workflow.actions.WarningAction;
-import org.jenkinsci.plugins.workflow.flow.FlowExecution;
 import org.jenkinsci.plugins.workflow.flow.FlowExecutionOwner;
 import org.jenkinsci.plugins.workflow.graph.FlowNode;
 import org.jenkinsci.plugins.workflow.steps.FlowInterruptedException;
@@ -94,26 +93,29 @@ public class BuildTriggerListener extends RunListener<Run<?,?>>{
                 BuildUpstreamCause buildUpstreamCause = (BuildUpstreamCause) cause;
                 Run<?, ?> upstream = buildUpstreamCause.getUpstreamRun();
                 if (upstream instanceof FlowExecutionOwner.Executable) {
-                    FlowExecutionOwner owner = ((FlowExecutionOwner.Executable) upstream).asFlowExecutionOwner();
-                    if (owner == null) {
-                        LOGGER.log(Level.FINE, () -> "Unable to update DownstreamBuildAction for " + upstream + " node " + buildUpstreamCause.getNodeId());
-                        continue;
+                    DownstreamBuildAction action = findMatchingAction(upstream, buildUpstreamCause.getNodeId());
+                    if (action == null) {
+                        action = new DownstreamBuildAction(buildUpstreamCause.getNodeId(), run.getParent());
+                        run.addAction(action);
                     }
+                    action.setBuild(run);
                     try {
-                        FlowExecution execution = owner.get();
-                        FlowNode node = execution.getNode(buildUpstreamCause.getNodeId());
-                        if (node == null) {
-                            LOGGER.log(Level.FINE, () -> "Unable to update DownstreamBuildAction for " + upstream + " node " + buildUpstreamCause.getNodeId());
-                            continue;
-                        }
-                        DownstreamBuildAction action = new DownstreamBuildAction(run);
-                        node.addOrReplaceAction(action);
-                        run.save();
+                        upstream.save();
+                        LOGGER.log(Level.WARNING, () -> "Saved DownstreamBuildAction for " + upstream + " node " + buildUpstreamCause.getNodeId());
                     } catch (IOException e) {
                         LOGGER.log(Level.FINE, e, () -> "Unable to update DownstreamBuildAction for " + upstream + " node " + buildUpstreamCause.getNodeId());
                     }
                 }
             }
         }
+    }
+
+    private DownstreamBuildAction findMatchingAction(Run<?, ?> run, String flowNodeId) {
+        for (DownstreamBuildAction action : run.getActions(DownstreamBuildAction.class)) {
+            if (action.getFlowNodeId().equals(flowNodeId)) {
+                return action;
+            }
+        }
+        return null;
     }
 }
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerListener.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerListener.java
@@ -87,35 +87,21 @@ public class BuildTriggerListener extends RunListener<Run<?,?>>{
         }
     }
 
-    private void updateDownstreamBuildAction(Run<?, ?> run) {
-        for (Cause cause : run.getCauses()) {
+    private void updateDownstreamBuildAction(Run<?, ?> downstream) {
+        for (Cause cause : downstream.getCauses()) {
             if (cause instanceof BuildUpstreamCause) {
                 BuildUpstreamCause buildUpstreamCause = (BuildUpstreamCause) cause;
                 Run<?, ?> upstream = buildUpstreamCause.getUpstreamRun();
                 if (upstream instanceof FlowExecutionOwner.Executable) {
-                    DownstreamBuildAction action = findMatchingAction(upstream, buildUpstreamCause.getNodeId());
-                    if (action == null) {
-                        action = new DownstreamBuildAction(buildUpstreamCause.getNodeId(), run.getParent());
-                        run.addAction(action);
-                    }
-                    action.setBuild(run);
+                    String flowNodeId = buildUpstreamCause.getNodeId();
+                    DownstreamBuildAction.getOrCreate(upstream, flowNodeId, downstream.getParent()).setBuild(downstream);
                     try {
                         upstream.save();
-                        LOGGER.log(Level.WARNING, () -> "Saved DownstreamBuildAction for " + upstream + " node " + buildUpstreamCause.getNodeId());
                     } catch (IOException e) {
                         LOGGER.log(Level.FINE, e, () -> "Unable to update DownstreamBuildAction for " + upstream + " node " + buildUpstreamCause.getNodeId());
                     }
                 }
             }
         }
-    }
-
-    private DownstreamBuildAction findMatchingAction(Run<?, ?> run, String flowNodeId) {
-        for (DownstreamBuildAction action : run.getActions(DownstreamBuildAction.class)) {
-            if (action.getFlowNodeId().equals(flowNodeId)) {
-                return action;
-            }
-        }
-        return null;
     }
 }
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepExecution.java
@@ -66,8 +66,8 @@ public class BuildTriggerStepExecution extends AbstractStepExecutionImpl {
     @Override
     public boolean start() throws Exception {
         String job = step.getJob();
-        Run<?, ?> run = getContext().get(Run.class);
-        Item item = Jenkins.get().getItem(job, run.getParent(), Item.class);
+        Run<?, ?> upstream = getContext().get(Run.class);
+        Item item = Jenkins.get().getItem(job, upstream.getParent(), Item.class);
         if (item == null) {
             throw new AbortException("No item named " + job + " found");
         }
@@ -78,11 +78,11 @@ public class BuildTriggerStepExecution extends AbstractStepExecutionImpl {
         }
 
         FlowNode node = getContext().get(FlowNode.class);
-        run.addAction(new DownstreamBuildAction(node.getId(), item));
+        DownstreamBuildAction.getOrCreate(upstream, node.getId(), item);
 
         List<Action> actions = new ArrayList<>();
-        actions.add(new CauseAction(new BuildUpstreamCause(getContext().get(FlowNode.class), run)));
-        actions.add(new BuildUpstreamNodeAction(node, run));
+        actions.add(new CauseAction(new BuildUpstreamCause(getContext().get(FlowNode.class), upstream)));
+        actions.add(new BuildUpstreamNodeAction(node, upstream));
 
         if (item instanceof ParameterizedJobMixIn.ParameterizedJob) {
             final ParameterizedJobMixIn.ParameterizedJob project = (ParameterizedJobMixIn.ParameterizedJob) item;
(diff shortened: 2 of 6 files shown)


src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerListener.java
```diff
@@ -106,13 +106,8 @@ private void updateDownstreamBuildAction(Run<?, ?> run) {
...
-                        DownstreamBuildAction downstreamAction = node.getPersistentAction(DownstreamBuildAction.class);
-                        if (downstreamAction == null) {
-                            // Should only happen for builds already in the queue when this plugin is updated to include DownstreamBuildAction.
-                            downstreamAction = new DownstreamBuildAction(run.getParent());
-                            node.addAction(downstreamAction);
-                        }
-                        downstreamAction.setBuild(run);
+                        DownstreamBuildAction action = new DownstreamBuildAction(run);
+                        node.addOrReplaceAction(action);
                         run.save();
```
jglick, line 111: Is the `run.save()` still needed?

jglick (approved):

Artmorse (approved):

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerListener.java
dwnusbaum: `WorkflowRun.save` + `CpsFlowExecution$ConverterImpl` will not flush the `FlowNodeStorage` as far as I can tell, so the `run.save` call seems pointless.

The situation seems complicated though. Mutation of the list returned by `FlowNode.getActions` calls `FlowNode.save` which calls `FlowActionStorage.saveActions`, and `CpsFlowExecution` routes that to `FlowNodeStorage.saveActions`. For `SimpleXStreamFlowNodeStorage` in most cases I think this will actually write the changes to disk, but not always. For `BulkFlowNodeStorage` it will just set a flag that something has been modified.

`CpsFlowExecution.saveOwner` and `FlowNodeStorage.storeNode(node, false)` would flush the `FlowNodeStorage`, but are not accessible here. I think that with the current code, if you use `build(wait: false, ...)`, and the upstream build completes before the downstream build starts, and you are running https://github.com/jenkinsci/workflow-cps-plugin/pull/807 or using `PERFORMANCE_OPTIMIZED`, then nothing will ever save the change to disk.

Not sure if there is a simple fix. We could maybe switch to storing this action along with the `FlowNode` IDs on `Run` instead. I guess this general issue applies to any attempts to modify actions on a `FlowNode` for a possibly-completed Pipeline run (probably very uncommon). We could introduce a new overload of `FlowActionStorage.saveActions` that forces the storage to be flushed if we care about the general case.

jglick: > We could maybe switch to storing this action along with the `FlowNode` IDs on `Run` instead.

See https://github.com/jenkinsci/pipeline-build-step-plugin/pull/13#issuecomment-1847847813. https://github.com/search?q=org%3Ajenkinsci%20DownstreamBuildAction&type=code shows no usages yet so it would be possible to do so without breaking anything. That may be better than trying to fix

> any attempts to modify actions on a `FlowNode` for a possibly-completed Pipeline run

though if we believe that to be unsupported (compare `onFinalized` w.r.t. the log file) then it would be better to explicitly throw an `IOException` saying so. (Whereas there are some established reasons why `<actions>` in `build.xml` might be modified after completion.)

Either way, it seems that

> the upstream build completes before the downstream build starts

deserves some test coverage too.

dwnusbaum: Yes I think I will switch to storing the data on the `Run`.

> though if we believe that to be unsupported (compare onFinalized w.r.t. the log file) then it would be better to explicitly throw an IOException saying so. 

Yeah I agree.

> Either way, it seems that
> > the upstream build completes before the downstream build starts
>
> deserves some test coverage too.

The current test does cover this case, it's just that we haven't picked up https://github.com/jenkinsci/workflow-cps-plugin/pull/807 yet and aren't using `PERFORMANCE_OPTIMIZED`, so it passes. I will try to expand the coverage though.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepRestartTest.java
```diff
@@ -64,6 +78,81 @@ public void restartBetweenJobs() throws Throwable {
         sessions.then(j -> j.jenkins.getItemByFullName("test1", FreeStyleProject.class).getBuildByNumber(1).delete());
     }
 
+    @Test
+    public void downstreamBuildActionUpstreamCompletesBeforeDownstreamStarts() throws Throwable {
+        sessions.then(j -> {
+            FreeStyleProject downstream = j.createFreeStyleProject("downstream");
+            downstream.setAssignedLabel(Label.parseExpression("agent"));
+            WorkflowJob upstream = j.jenkins.createProject(WorkflowJob.class, "upstream");
+            upstream.addProperty(new DurabilityHintJobProperty(FlowDurabilityHint.PERFORMANCE_OPTIMIZED)); // TODO: Unnecessary once we pick up https://github.com/jenkinsci/workflow-cps-plugin/pull/807.
```
dwnusbaum, line 87: Not strictly relevant anymore, but seems best to test the lazier implementation.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerListener.java
```diff
@@ -94,31 +93,29 @@ private void updateDownstreamBuildAction(Run<?, ?> run) {
...
-                        }
-                        DownstreamBuildAction downstreamAction = node.getPersistentAction(DownstreamBuildAction.class);
-                        if (downstreamAction == null) {
-                            // Should only happen for builds already in the queue when this plugin is updated to include DownstreamBuildAction.
-                            downstreamAction = new DownstreamBuildAction(run.getParent());
-                            node.addAction(downstreamAction);
-                        }
-                        downstreamAction.setBuild(run);
-                        run.save();
+                        upstream.save();
```
dwnusbaum, line 103: Comment this out and `downstreamBuildActionUpstreamCompletesBeforeDownstreamStarts` will fail.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerListener.java
dwnusbaum: See https://github.com/jenkinsci/pipeline-build-step-plugin/pull/128/commits/cd63d16e2d6a3b5aa4736c73ff1e2a87b1e7b44a.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepRestartTest.java
jglick: Why not just update the `bom` now to pick up https://github.com/jenkinsci/workflow-cps-plugin/pull/807?

src/test/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepRestartTest.java
```diff
@@ -64,6 +78,81 @@ public void restartBetweenJobs() throws Throwable {
...
+
+    @Test
+    public void downstreamBuildActionMultipleBuilds() throws Throwable {
+        sessions.then(j -> {
+            FreeStyleProject downstream = j.createFreeStyleProject("downstream");
+            WorkflowJob upstream = j.jenkins.createProject(WorkflowJob.class, "upstream");
+            upstream.setDefinition(new CpsFlowDefinition("build(job: 'downstream'); build(job: 'downstream')", true));
+            WorkflowRun upstreamRun = j.buildAndAssertSuccess(upstream);
+            Map<String, DownstreamBuildAction> actions = upstreamRun.getActions(DownstreamBuildAction.class).stream()
+                    .collect(Collectors.toMap(a -> a.getFlowNodeId(), a -> a));
```
jglick, line 128: I find it surprising that there are multiple `DownstreamBuildAction`s. Typically actions are singletons, with substructure as needed (compare `CauseAction.causes`).

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerListener.java
```diff
@@ -94,31 +93,29 @@ private void updateDownstreamBuildAction(Run<?, ?> run) {
...
                 Run<?, ?> upstream = buildUpstreamCause.getUpstreamRun();
                 if (upstream instanceof FlowExecutionOwner.Executable) {
-                    FlowExecutionOwner owner = ((FlowExecutionOwner.Executable) upstream).asFlowExecutionOwner();
-                    if (owner == null) {
-                        LOGGER.log(Level.FINE, () -> "Unable to update DownstreamBuildAction for " + upstream + " node " + buildUpstreamCause.getNodeId());
-                        continue;
+                    DownstreamBuildAction action = findMatchingAction(upstream, buildUpstreamCause.getNodeId());
+                    if (action == null) {
+                        action = new DownstreamBuildAction(buildUpstreamCause.getNodeId(), run.getParent());
+                        run.addAction(action);
```
jglick, line 99: Did you mean to add this to `upstream`? IIUC `run` is the downstream build (perhaps best to rename it for clarity)?

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerListener.java
```diff
@@ -94,31 +93,29 @@ private void updateDownstreamBuildAction(Run<?, ?> run) {
...
-                        DownstreamBuildAction downstreamAction = node.getPersistentAction(DownstreamBuildAction.class);
-                        if (downstreamAction == null) {
-                            // Should only happen for builds already in the queue when this plugin is updated to include DownstreamBuildAction.
-                            downstreamAction = new DownstreamBuildAction(run.getParent());
-                            node.addAction(downstreamAction);
-                        }
-                        downstreamAction.setBuild(run);
-                        run.save();
+                        upstream.save();
+                        LOGGER.log(Level.WARNING, () -> "Saved DownstreamBuildAction for " + upstream + " node " + buildUpstreamCause.getNodeId());
```
jglick, line 104:
```suggestion
                        LOGGER.fine(() -> "Saved DownstreamBuildAction for " + upstream + " node " + buildUpstreamCause.getNodeId());
```
(use `LoggerRule` from tests as needed)

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepExecution.java
```diff
@@ -66,7 +66,8 @@ public BuildTriggerStepExecution(BuildTriggerStep step, @NonNull StepContext con
     @Override
     public boolean start() throws Exception {
         String job = step.getJob();
-        Item item = Jenkins.get().getItem(job, getContext().get(Run.class).getParent(), Item.class);
+        Run<?, ?> run = getContext().get(Run.class);
```
jglick, line 69: perhaps rename to `upstream` for clarity

src/test/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepRestartTest.java
```diff
@@ -64,6 +78,81 @@ public void restartBetweenJobs() throws Throwable {
...
+            assertThat(action.getBuild(), equalTo(downstream.getLastBuild()));
+        });
+    }
+
+    @Test
+    public void downstreamBuildActionMultipleBuilds() throws Throwable {
+        sessions.then(j -> {
+            FreeStyleProject downstream = j.createFreeStyleProject("downstream");
+            WorkflowJob upstream = j.jenkins.createProject(WorkflowJob.class, "upstream");
+            upstream.setDefinition(new CpsFlowDefinition("build(job: 'downstream'); build(job: 'downstream')", true));
```
jglick, line 125: or more idiomatically
```suggestion
            upstream.setDefinition(new CpsFlowDefinition("build 'downstream'; build 'downstream'", true));
```

src/test/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepRestartTest.java
```diff
@@ -64,6 +78,81 @@ public void restartBetweenJobs() throws Throwable {
...
     }
 
+    @Test
+    public void downstreamBuildActionUpstreamCompletesBeforeDownstreamStarts() throws Throwable {
+        sessions.then(j -> {
+            FreeStyleProject downstream = j.createFreeStyleProject("downstream");
+            downstream.setAssignedLabel(Label.parseExpression("agent"));
+            WorkflowJob upstream = j.jenkins.createProject(WorkflowJob.class, "upstream");
+            upstream.addProperty(new DurabilityHintJobProperty(FlowDurabilityHint.PERFORMANCE_OPTIMIZED)); // TODO: Unnecessary once we pick up https://github.com/jenkinsci/workflow-cps-plugin/pull/807.
+            upstream.setDefinition(new CpsFlowDefinition("build(job: 'downstream', wait: false)", true));
```
jglick, line 88: idiomatically
```suggestion
            upstream.setDefinition(new CpsFlowDefinition("build job: 'downstream', wait: false", true));
```

src/test/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepRestartTest.java
dwnusbaum: Because this was trivial and didn't require me to look at anything else. Looking now at dependency versions it does seem that we have the same Jenkins minimum version requirements in both plugins so I can update the BOM.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepRestartTest.java
dwnusbaum: I was thinking of Git's `BuildData`, but sure I can refactor everything if desired.

jglick: Do not have a strong feeling about it, though I think the singleton style is preferred as a rule. `InterruptedBuildAction.causes` is another example.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/DownstreamBuildAction.java
```diff
@@ -6,47 +6,80 @@
...
-    private final String jobFullName;
-    private Integer buildNumber;
+public final class DownstreamBuildAction extends InvisibleAction {
+    private final Map<String, DownstreamBuild> downstreamBuilds = new LinkedHashMap<>();
 
-    DownstreamBuildAction(Item job) {
-        this.jobFullName = job.getFullName();
+    public static @NonNull DownstreamBuild getOrCreate(@NonNull Run<?, ?> run, @NonNull String flowNodeId, @NonNull Item job) {
+        DownstreamBuildAction downstreamBuildAction;
+        synchronized (DownstreamBuildAction.class) {
```
dwnusbaum, line 25: Unnecessary in the normal case since the relevant part of `BuildTriggerStepExecution` runs synchronously on the CPS VM thread, but if there are queued triggered builds when updating Jenkins to pick up this PR or when Jenkins crashes then there is a possibility of race conditions leading to multiple instances of `DownstreamBuildAction` without some kind of synchronization.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/DownstreamBuildAction.java
```diff
@@ -6,53 +6,80 @@
...
- * Tracks a downstream build triggered by the {@code build} step, as well as the {@link FlowNode#getId} of the step.
+ * Tracks downstream builds triggered by the {@code build} step, as well as the {@link FlowNode#getId} of the step.
  *
  * @see BuildUpstreamCause
  */
 public final class DownstreamBuildAction extends InvisibleAction {
-    private final String flowNodeId;
-    private final String jobFullName;
-    private Integer buildNumber;
+    private final Map<String, DownstreamBuild> downstreamBuilds = new LinkedHashMap<>();
```
jglick, line 21: Works, but note that this will look a bit funny in XStream format. I would suggest using an `ArrayList` and if any indexing is necessary (unlikely, since median length is 1) do that in memory.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/DownstreamBuildAction.java
jglick: May as well keep it, does not add much clutter.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepRestartTest.java
dwnusbaum: See https://github.com/jenkinsci/pipeline-build-step-plugin/pull/128/commits/ec1d5f8a45516a234b233af384f2b299a724492d. FWIW after making the change I think if you have a case like this where the action is not created deterministically at build start (although I guess we could do that) _and_ you need to update the set of internal items dynamically using one action per internal item seems much more straightforward.

jglick (approved):

jglick:
Not sure if https://github.com/jenkinsci/pipeline-build-step-plugin/pull/128/checks?check_run_id=19658148636 is related; maybe independently flaky?

dwnusbaum:
> Not sure if https://github.com/jenkinsci/pipeline-build-step-plugin/pull/128/checks?check_run_id=19658148636 is related; maybe independently flaky?

I am not sure. It passes locally but has failed twice in CI in this PR, once on Windows before the BOM update, and once on Linux after the BOM update. In both cases the problem seems to be that the downstream build never finishes (it seems to get stuck on a `semaphore` step even after a call to `SemaphoreStep.success`).  

`BuildTriggerStepRestartTest.restartBetweenJobs` also failed in an earlier build.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/DownstreamBuildAction.java
dwnusbaum: https://github.com/jenkinsci/pipeline-build-step-plugin/pull/128/commits/339adf57162379e9d3ed75c78c02c7e483f7ce32

jglick (approved):

PR #146, opened 2025-01-25
De-flake `BuildTriggerStepTest.waitForStart`

basil:
When running with `-Dorg.jvnet.hudson.test.RemainingActivityListener.fatal=true` I got the following error (fixed in this PR):

```
[ERROR] org.jenkinsci.plugins.workflow.support.steps.build.BuildTriggerStepTest.waitForStart -- Time elapsed: 0.993 s <<< FAILURE!
java.lang.AssertionError: ds #1 still seems to be running, which could break deletion of log files or metadata
        at org.jvnet.hudson.test.RemainingActivityListener.onTearDown(RemainingActivityListener.java:62)
        at org.jvnet.hudson.test.JenkinsRule.after(JenkinsRule.java:502)
        at org.jvnet.hudson.test.JenkinsRule$1.evaluate(JenkinsRule.java:677)
        at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
        at java.base/java.lang.Thread.run(Thread.java:1583)
```

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepTest.java
@@ -454,10 +454,12 @@ public class BuildTriggerStepTest {
     }
 
     @Test public void waitForStart() throws Exception {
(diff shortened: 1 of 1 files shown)


jglick (approved):

jglick:
(`developer` label upon request)

PR #151, opened 2025-02-25
Update LabelAction messages for Build and WaitForBuild steps

stuartrowe:
The label for the `build` step should differentiate between whether the step is waiting for the scheduled build to complete. Similarly the `waitForBuild` step label should reference the build it's waiting on.

image
image

### Testing done

Ran local pipelines that called `build` with a combination of `wait=true`, `wait=false` and `waitForStart=true` along with calling the `waitForBuild` step. Confirmed that labels associated with the step nodes was updated accordingly.

### Submitter checklist
- [X] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [X] Ensure that the pull request title represents the desired changelog entry
- [X] Please describe what you did
- [ ] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepExecution.java
@@ -88,7 +88,11 @@ public class BuildTriggerStepExecution extends AbstractStepExecutionImpl {
             final ParameterizedJobMixIn.ParameterizedJob project = (ParameterizedJobMixIn.ParameterizedJob) item;
             getContext().get(TaskListener.class).getLogger().println("Scheduling project: " + ModelHyperlinkNote.encodeTo(project));
 
-            node.addAction(new LabelAction(Messages.BuildTriggerStepExecution_building_(project.getFullDisplayName())));
+            if (!step.getWait() || step.getWaitForStart()) {
+                node.addAction(new LabelAction(Messages.BuildTriggerStepExecution_scheduling_(project.getFullDisplayName())));
+            } else {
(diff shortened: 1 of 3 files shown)


src/main/resources/org/jenkinsci/plugins/workflow/support/steps/build/Messages.properties
```diff
@@ -28,7 +28,9 @@ BuildTriggerStep.no_job_configured=No job configured
 BuildTriggerStep.cannot_find=No such job {0}
 BuildTriggerStep.unsupported=Building a {0} is not supported
 BuildTriggerStepExecution.building_=Building {0}
+BuildTriggerStepExecution.scheduling_=Scheduling {0}
```
jglick, line 31: BTW the trailing `_` in the original key was probably a mistake; no need to copy it.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepExecution.java
```diff
@@ -115,6 +119,12 @@ public boolean start() throws Exception {
             Queue.Task task = (Queue.Task) item;
             getContext().get(TaskListener.class).getLogger().println("Scheduling item: " + ModelHyperlinkNote.encodeTo(item));
             node.addAction(new LabelAction(Messages.BuildTriggerStepExecution_building_(task.getFullDisplayName())));
```
jglick, line 121:
```suggestion
```
?

jglick (commented):
Apparently duplicated line. Change otherwise looks reasonable.

jglick (approved):

PR #161, opened 2025-06-27
Avoid loading the downstream build in `DownstreamFailureCause.getShortDescription` and `.print`

dwnusbaum:
The deadlock originally reported in https://github.com/jenkinsci/opentelemetry-plugin/issues/700 was partially caused by the fact that `DownstreamFailureCause.getShortDescription` triggered build loading. This is not _wrong_, but it seems unnecessarily risky given that `FlowInterruptedException.getCauses` can be processed in various contexts. I think we should instead persist the relevant attributes directly. Compare with [`Cause.UpstreamCause`](https://github.com/jenkinsci/jenkins/blob/7beaf2ee4c364c9e5f42426812e5aca360db0fb0/core/src/main/java/hudson/model/Cause.java#L160-L161).

### Submitter checklist
- [ ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [ ] Ensure that the pull request title represents the desired changelog entry
- [ ] Please describe what you did
- [ ] Ensure you have provided tests that demonstrate the feature works or the issue is fixed


src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/DownstreamFailureCause.java
```diff
@@ -37,35 +40,60 @@ public final class DownstreamFailureCause extends CauseOfInterruption {
...
+        jobFullName = downstream.getParent().getFullName();
+        buildNumber = downstream.getNumber();
+        url = downstream.getUrl();
+        result = downstream.getResult();
     }
 
     public @CheckForNull Run<?, ?> getDownstreamBuild() {
-        return Run.fromExternalizableId(id);
+        if (id != null) {
+            return Run.fromExternalizableId(id);
```
jglick, line 60: (hide WS)

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/DownstreamFailureCause.java
```diff
@@ -37,35 +40,60 @@ public final class DownstreamFailureCause extends CauseOfInterruption {
...
+            Run<?, ?> downstream = getDownstreamBuild();
+            if (downstream != null) {
+                // encodeTo(Run) calls getDisplayName, which does not include the project name.
+                description = description(ModelHyperlinkNote.encodeTo("/" + downstream.getUrl(), downstream.getFullDisplayName()), downstream.getResult());
+            } else {
+                description = "Downstream build was not stable (propagate: false to ignore)";
+            }
         } else {
-            description = "Downstream build was not stable (propagate: false to ignore)";
+            description = description(ModelHyperlinkNote.encodeTo("/" + url, jobFullName + " #" + buildNumber), result);
```
jglick, line 77: Note that this is swapping `fullDisplayName` for `fullName`, so ` » ` ⇒ `/` as well as ignoring any display names set on the job and/or folders.

jglick (approved):

PR #177, opened 2026-02-19
Fix: Prevent `IllegalStateException` When Downstream Builds Are Deleted

bingshiao:
### Summary
This PR fixes an `IllegalStateException` thrown when a downstream build is deleted after its upstream Pipeline has already completed or become broken. The issue appears when older builds are cleaned up by the "Discard old builds" option. See https://github.com/jenkinsci/pipeline-build-step-plugin/issues/253 for more details.

The root cause is that `BuildTriggerListener.onDeleted()` calls `StepContext.onFailure()` **unconditionally**, even when the associated `CpsStepContext` is no longer able to accept completion signals (i.e., the upstream execution has already finished).

### Root Cause
When a downstream build is deleted:
1. `BuildTriggerListener.onDeleted()` schedules a timer task that calls `CpsStepContext.onFailure()`.
2. This eventually triggers a load of the upstream `CpsFlowExecution`.
3. If the upstream execution is already completed or broken, `CpsFlowExecution.onLoad()` sets `programPromise` to a failed future.
4. The error propagates back as an `IllegalStateException`, generating noisy and misleading warnings in the logs.

This happens specifically during build cleanup events, such as when Jenkins deletes old runs automatically.

### Fix
`BuildTriggerListener.onDeleted()` now checks `stepContext.isReady()` before attempting to call `onFailure()`, matching the logic already used in `onStarted()`.

### Why This Works
`isReady()` ensures the upstream execution is still alive and accepting completion callbacks. Guarding the call prevents invalid state transitions and eliminates the `IllegalStateException` warnings without changing any functional behavior for valid cases.

### Submitter checklist
- [x] Ensure you have provided tests that demonstrate the feature works or the issue is fixed

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerListener.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerListener.java
@@ -83,7 +83,12 @@ public class BuildTriggerListener extends RunListener<Run<?,?>>{
     @Override
     public void onDeleted(final Run<?,?> run) {
         for (final BuildTriggerAction.Trigger trigger : BuildTriggerAction.triggersFor(run)) {
-            Timer.get().submit(() -> trigger.context.onFailure(new AbortException(run.getFullDisplayName() + " was deleted")));
+            StepContext stepContext = trigger.context;
+            if (stepContext != null && stepContext.isReady()) {
+                Timer.get().submit(() -> stepContext.onFailure(new AbortException(run.getFullDisplayName() + " was deleted")));
+            } else {
+                LOGGER.log(Level.FINE, "{0} unavailable when {1} was deleted", new Object[] {stepContext, run});
+            }
         }
     }
 


jglick:
> Of the upstream execution is already completed [… the] error propagates back as an `IllegalStateException`, generating noisy and misleading warnings

Is there some way to reproduce that? https://github.com/jenkinsci/pipeline-build-step-plugin/blob/7d122e63010dcd0f9f56c793c190c030ffbc69ed/src/test/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepTest.java#L121-L123 (from #2) does not.

fherrero:
Hello,

This fix helped us with an error we were having. We have two jobs used for functional testing. The Test Plan job retrieves a number of test cases and launch a Test case job for each one. Jenkins is configured to only retain the last 30 items of a job. If the Test Plan happened to have more than 30 test cases, the Test Plan job failed when one of the already finished test cases is deleted, even though that test case is no longer needed and had already finished its execution. That prevented the rest of test cases to execute.

By compiling the plugin with this fix, our problem was solved. So thank you for the fix and it would be great if it could be merged in a future version of the plugin.

Kind Regards,

TimotheusBachinger:
hi all, 
We have rebuilt and deployed the plugin with that patch and our logs are not flooded anymore with `IllegalStateException` - so it works pretty good for us.
Is there a chance to get this merged upstream?

Regards
Timi

jglick:
I am still awaiting some information on how to reproduce the problem this change purportedly fixes. (Preferably in the form of new test coverage, but any complete, self-contained steps to reproduce from scratch would do.)

jglick:
Similar fix in https://github.com/jenkinsci/pipeline-build-step-plugin/pull/254 with test.

